PR #1, opened 2017-11-15
Move secondary-analysis-python code from skylab to here

dshiga:
As part of repo reorganization, I'd like to move this code here from https://github.com/HumanCellAtlas/skylab/tree/master/docker/secondary-analysis-python.

I haven't changed it, other than to rename secondary_analysis_python to pipeline_tools, which also required a small change to the Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+FROM python:2.7
+
+LABEL maintainer="Mint Team <mintteam@broadinstitute.org>" \
+  software="Python" \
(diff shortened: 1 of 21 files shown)


Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM python:2.7
+
+LABEL maintainer="Mint Team <mintteam@broadinstitute.org>" \
+  software="Python" \
+  description="Python2.7 with some useful libraries used for secondary-analysis-python application"
```
rexwangcc: future AC here: secondary-analysis-python rename to pipeline-tools

PR #3, opened 2017-11-17
Add gcs_utils

dshiga:
Move gcs_utils and associated tests from lira to here. See associated PR in Lira here: https://github.com/HumanCellAtlas/lira/pull/35


pipeline_tools/gcs_utils.py
```diff
@@ -0,0 +1,108 @@
...
+def parse_bucket_blob_from_gs_link(path):
+    """Utility to split a google storage path into bucket + blob name.
+
+    :param str path: A string of google cloud storage path (must have gs:// prefix)
+    :return str: A string of bucket name.
+    :return str: A string of blob name
+    """
+    if not path.startswith('gs://'):
+        raise ValueError('%s path is not a valid link')
+    parts = path.split('/', 3)
```
ambrosejcarr: I had no idea this took a numerical parameter. You have changed my life.

pipeline_tools/gcs_utils.py
```diff
@@ -0,0 +1,108 @@
...
+    return bucket, blob
+
+
+def download_to_buffer(blob):
+    """Return a bytes file-like object readable by requests and REST APIs
+
+    :param google.cloud.storage.Blob blob: google storage blob
+    :return _io.BufferedIOBase: readable file object
+    """
+    string_buffer = BytesIO()
```
ambrosejcarr: I always try to name these `bytes_buffer` because `io.BufferedIOBase` doesn't clearly indicate that this is a bytes object you're getting.

pipeline_tools/gcs_utils.py
```diff
@@ -0,0 +1,108 @@
...
+        gcs_client.storage_client
+    authenticated_gcs_client = gcs_client.storage_client
+
+    bucket = authenticated_gcs_client.bucket(bucket_name)
+    blob = bucket.blob(source_blob_name)
+    logging.debug('bucket: {0}, blob: {1}'.format(bucket_name, source_blob_name))
+    return download_to_buffer(blob)
+
+
+class LazyProperty(object):
```
ambrosejcarr: @rexwangcc Is this replaceable with python's `functools.lru_cache` with the max_size parameter set to 1?

rexwangcc: I did a quick research about the [functools implementation](https://github.com/python/cpython/blob/60a376cf0023d1070329d0e861a5596637ff3275/Lib/functools.py#L399) of LRU cache. I think it is possible we could replace the current decorator with the lru_cache decorator. To do this, we need to do one of the following two things: 
1. We could still use the GoogleCloudStorageClient class, but we need to implement the `__hash__` method, to enable lru_cache to work for the client, we might need to change the way that passing credentials to the client, those lists are not hashable. In short, in this way, we make the whole GoogleCloudStorageClient class hashable, and decorate the download_gcs_blob with lru_cache, since the max size is 1 and it is thread-safe, it will always hit the cache. 
2. A neater way requires more refactoring work: refactor the code and remove the GoogleCloudStorageClient class. Instead, we just need to use storage_client function and decorate the download_gcs_blob. Since we are using storage_client to store the information of the credentials, this will also ACT-AS the lazy property, and will only throw errors during the runtime.

For the second way, I tested it a little bit, codes are shown as below(alert me if there are any credentials I put in the code):
```python
from google.cloud import storage
from google.oauth2 import service_account
from io import BytesIO
from functools import lru_cache

def download_to_buffer(blob):
    string_buffer = BytesIO()
    blob.download_to_file(string_buffer)
    string_buffer.seek(0)
    return string_buffer

@lru_cache(maxsize=1)
def download_gcs_blob(bucket_name, source_blob_name):
    authenticated_gcs_client = storage_client()
    bucket = authenticated_gcs_client.bucket(bucket_name)
    blob = bucket.blob(source_blob_name)
    return download_to_buffer(blob)


def storage_client(key_location='/etc/secondary-analysis/bucket-reader-key.jon', scopes=['https://www.googleapis.com/auth/devstorage.read_only']):
    credentials = service_account.Credentials.from_service_account_file(
        key_location, scopes=scopes)
    client = storage.Client(credentials=credentials, project=credentials.project_id)
    return client

for i in range(40):
    wdlIO = download_gcs_blob("broad-dsde-mint-dev-teststorage", "prepare_and_analyze_mock_smartseq2.wdl")

download_gcs_blob.cache_info()
"""Out[2]: CacheInfo(hits=39, misses=1, maxsize=1, currsize=1)"""
```

The result looks reasonable, if we gonna use this build-in decorator, we could benefit from the official support for the decorator as well as the fast velocity of C-implementation of the LRU cache algorithm, and this could be more scalable than our current lazy property. But whether to change or not depends on where we still want to download wdls from, Github or buckets. Overall, thank you very much for this idea! @ambrosejcarr

dshiga: Interesting idea. It would be nice to simplify this code. Let's make a separate ticket for this.

pipeline_tools/gcs_utils.py
dshiga: good idea

PR #5, opened 2017-12-14
Cw adapt to v4 metadata 408

rexwangcc:
This PR is intended to embrace the v4(4.6.1 for now) metadata schema released in this quarter. 

This PR mainly related to the interaction between the Purple Box and Green Box, there is another PR about the connection between Blue Box and Green Box: https://github.com/HumanCellAtlas/lira/pull/45

I did the following things in this PR:

1. Update the **ss2/adapter.wdl**:
	- `sample_id = assay_json['sample_id']` -> `sample_id = assay_json['has_input']`
	- `fastq_1_name = assay_json['seq']['lanes'][0]['r1']` -> `fastq_1_name = assay_json['content']['seq']['lanes'][0]['r1']`
	- `fastq_2_name = assay_json['seq']['lanes'][0]['r2']` -> `fastq_2_name = assay_json['content']['seq']['lanes'][0]['r2']`
	- `from pipeline_tools import utils` -> `from pipeline_tools import dcp_utils`
	- `utils.get_manifest_files()` -> `dcp_utils.get_manifest_files()`
	- `utils.get_file_by_uuid()` -> `dcp_utils.get_file_by_uuid`
2. Update the **10x/adapter.wdl**:
	- `sample_id = assay_json['sample_id']` -> `sample_id = assay_json['has_input']`
	- `lanes = assay_json['seq']['lanes']` -> `lanes = assay_json['content']['seq']['lanes']`
	- `from pipeline_tools import utils` -> `from pipeline_tools import dcp_utils`
	- `utils.get_manifest_files()` -> `dcp_utils.get_manifest_files()`
	- `utils.get_file_by_uuid()` -> `dcp_utils.get_file_by_uuid`
3. Update the **submit.wdl** and the **pipeline-tools/create_analysis_json.py**:
	- add `create_core(type, schema_version)` to pipeline-tools/create_analysis_json.py for creating `core` object.
	- add unit tests
	- fix `setup.py`
4. Fix the **pipeline-tools/create_envelop.py** function:
	- add `schema_version = analysis_json['metadata_schema']` to get the schema version
	- add `'core': create_core(type='file', schema_version=schema_version)` in output_ref to match the new schema
5. Update the **docker image** `humancellatlas/pipeline-tools:0.1.4`:
	- add `.dockerignore` file
	- update the version of `hca-cli`
	- rebuild and push the docker: `humancellatlas/pipeline-tools:0.1.5`
6. Add examples:
	- add 10x/adapter_example_static_v4.json
[...]

PR #6, opened 2017-12-17
Handle parsing subworkflows when creating analysis.json 

samanehsan:
When parsing the metadata for workflow calls in `create_analysis_json.py` check if a call is a subworkflow and parse its calls recursively. This is needed in order to run the optimus and new smartseq2 adapter wdls, since those pipelines include subworkflows.

Note: The `humancellatlas/pipeline-tools:0.1.4` docker image will need to be updated to include these changes, so it would be easier to merge this into #5 if possible, since it makes changes to this script as well.


rexwangcc:
Thanks for this change, which is really useful, since the dcp int test works now, the #5 is ready for review, and we could work together to merger these 2 PRs.

PR #7, opened 2017-12-18
Optimus adapter wdl

samanehsan:
Add adapter wdl for Optimus pipeline that is compatible with the v4.6.1 metadata schema. Needs changes in #5 and #6 merged first.


adapter_pipelines/Optimus/adapter_example_bundle_specific_v4.json
```diff
@@ -0,0 +1,4 @@
+{
+  "AdapterOptimus.bundle_uuid": "52a56d80-5d6b-4065-8931-2cd666951397",
```
samanehsan: This needs to be updated to a v4.6.1 bundle for 10x data

adapter_pipelines/Optimus/adapter.wdl
```diff
@@ -0,0 +1,239 @@
...
+      docker: "humancellatlas/pipeline-tools:0.1.5"
+    }
+
+    output {
+      File inputs = "inputs.tsv"
+    }
+}
+
+task outputs_for_submit {
+    Array[Array[File]] tag_gene_exon_log
```
rexwangcc: This is related to an Elastc ticket #487 that requires further discussion as a team, but for now, I believe it is good to be merged in.

adapter_pipelines/Optimus/adapter_example_bundle_specific_v4.json
rexwangcc: The only 10x_v2 data bundle in v4.6.1 schema we have is the Human Spleen data, we could update this with that bundle uuid, however, we need to make sure it is suitable to expose that data bundle id to the public here since it's a "production data". What do you think? @kbergin

samanehsan: To summarize our offline discussion, it's okay to use the spleen data bundle here but there should be a bundle with 10x demo data we can use for testing instead. I'll send in a follow-up PR to change this once we confirm what that bundle id is.

samanehsan:
Thanks for the review @rexwangcc!

PR #8, opened 2017-12-19
Add travis unit tests

dshiga:
Added a `.travis.yml` to run the unit tests in both Python 2 and 3.

Fixed a small Python 3 issue with importing from create_analysis_json. A dot is needed as a prefix for that import in 3 (either way works for 2).

Also removed test.sh. It was broken because test-requirements.txt is not installed in the docker image. We could theoretically install test-requirements.txt in the docker image in order to get unit tests running there, but it seems silly to . However, there are advantages to not running the tests in docker. It's a little faster to test if you don't have to build the docker image. Not using docker also let's us easily test against both Python 2 and 3 in Travis. The docker image itself will still get tested as part of integration tests.


.travis.yml
```diff
@@ -0,0 +1,6 @@
+language: python
+python:
+  - "2.7"
+  - "3.6"
+install: pip install -r requirements.txt -r test-requirements.txt
+script: python -m unittest discover -v
```
rexwangcc: The same as I commented in another PR: 
```yaml
notifications:
  slack:
    on_success: change
    on_failure: always
    secure: xxx-xxx-xxx-xxx
```

PR #9, opened 2017-12-21
Cw adapt to purple auth endpoints 146

rexwangcc:
In this PR, I refactor code of pipeline-tools to support communicating with the new authentication endpoints in ingest service. Also, add unit tests for dcp_utils.


pipeline_tools/dcp_utils.py
```diff
@@ -59,3 +59,52 @@ def get_manifest_files(bundle_uuid, bundle_version, dss_url, timeout_seconds, re
         'name_to_meta': name_to_meta,
         'url_to_name': url_to_name
     }
+
+
+def get_auth_token(url="https://danielvaughan.eu.auth0.com/oauth/token",
+                   client_id="Zdsog4nDAnhQ99yiKwMQWAPc2qUDlR99",
+                   client_secret="t-OAE-GQk_nZZtWn-QQezJxDsLXmU7VSzlAh9cKW5vb87i90qlXGTvVNAjfT9weF",
```
dshiga: Shouldn't this be secret @rexwangcc and @danielvaughan? Do we need a new secret now that this one has been exposed?

danielvaughan: The secret gives you the ability to request a token. However, anyone with a Google account can also get a token. At the moment there is no authorisation and the token is purely used to enable us to identify authenticated users and fill in owner field for the submission envelope in the database. The token that the secret gives you is a catch-all as it allows us to require the endpoint is sent and token and allows us to put something in the database for submission created without logging in via google. We also use the same token for creating submission through the previous UI. No one can do anything with the secret that they could do not do without. The secret can also revoked at any time should that change.

PR #10, opened 2017-12-21
Fix setup.py

dshiga:
PR #8 moved the tests dir to underneath pipeline_tools. We were specifying in setup.py to install a top level tests dir that no longer exists. This PR fixes that.


jsotobroad:
👍  The change makes sense but I'm a little confused on how the PR that moved the files pass its tests?  (Sorry I don't have much context here)

dshiga:
Yeah, good question, the setup.py only gets used when something else is installing pipeline-tools as a dependency. So unit tests for this repo weren't affected by this bug.

jsotobroad:
Ahh i c, thanks!

PR #12, opened 2018-01-05
Add SmartSeq2SingleSample adapter WDL

ambrosejcarr:
This PR adds an adapter WDL and the necessary input files to run the SmartSeq2SingleSample pipeline. 

The notebook will be removed prior to merging, but allows a user to test the wrapper code against Cromwell.

### DIFF
--- a/adapter_pipelines/ss2_single_sample/adapter.wdl
+++ b/adapter_pipelines/ss2_single_sample/adapter.wdl
@@ -1,0 +1,212 @@
+import "ss2_single_sample.wdl" as ss2
+import "submit.wdl" as submit_wdl
+
+
+task GetInputs {
+  String bundle_uuid
+  String bundle_version
+  String dss_url
+  Int retry_seconds
+  Int timeout_seconds
+
+  command <<<
+    python <<CODE
+    from pipeline_tools import dcp_utils
+
+    # Get bundle manifest
+    uuid = "${bundle_uuid}"
+    version = "${bundle_version}"
+    dss_url = "${dss_url}"
+    retry_seconds = ${retry_seconds}
+    timeout_seconds = ${timeout_seconds}
+    print("Getting bundle manifest for id {0}, version {1}".format(uuid, version))
+    manifest_files = dcp_utils.get_manifest_files(uuid, version, dss_url, timeout_seconds, retry_seconds)
+
+    print("Downloading assay.json")
+    assay_json_uuid = manifest_files["name_to_meta"]["assay.json"]["uuid"]
+    assay_json = dcp_utils.get_file_by_uuid(assay_json_uuid, dss_url)
+
+    sample_id = assay_json["has_input"]
+    fastq_1_name = assay_json["content"]["seq"]["lanes"][0]["r1"]
+    fastq_2_name = assay_json["content"]["seq"]["lanes"][0]["r2"]
+    fastq_1_url = manifest_files["name_to_meta"][fastq_1_name]["url"]
+    fastq_2_url = manifest_files["name_to_meta"][fastq_2_name]["url"]
+
+    print("Creating input map")
+    with open("inputs.tsv", "w") as f:
+        f.write("fastq_1\tfastq_2\tsample_id\n")
+        f.write("{0}\t{1}\t{2}\n".format(fastq_1_url, fastq_2_url, sample_id))
+    print("Wrote input map")
+    CODE
+  >>>
+  runtime {
+    docker: "humancellatlas/pipeline-tools:0.1.5"
+  }
+  output {
+    Object inputs = read_object("inputs.tsv")
+  }
+}
+
+workflow AdapterSmartSeq2SingleCell{
+  # variable parameters provided by the notification sent by the data storage service
+  String bundle_uuid
+  String bundle_version
+
+  # fixed parameters
+  File gtf_file
+  File genome_ref_fasta
+  File rrna_intervals
+  File gene_ref_flat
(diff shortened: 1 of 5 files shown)


adapter_pipelines/ss2_single_sample/dependencies.json
```diff
@@ -0,0 +1,9 @@
+{
```
samanehsan: Is this dependencies file only used by the ipython notebook?

adapter_pipelines/ss2_single_sample/adapter.wdl
```diff
@@ -0,0 +1,212 @@
...
+  runtime {
+    docker: "humancellatlas/pipeline-tools:0.1.5"
+  }
+  output {
+    Object inputs = read_object("inputs.tsv")
+  }
+}
+
+workflow AdapterSmartSeq2SingleCell{
+  # variable parameters from blue
```
samanehsan: I think this could be more descriptive here, like "variable parameters from the notification sent by the data storage service"

adapter_pipelines/ss2_single_sample/dependencies.json
ambrosejcarr: I imagine that lira will need to be told which wdls to import, so I think it will be used by lira as well.

rexwangcc: Refer to https://github.com/HumanCellAtlas/lira/blob/2362b238d706cfec2d53a26de6bbbbf4945aa8f3/lira/test/data/config.json#L21, I assume if we update the config file of lira in that way, we may don't need this dependenceis.json anymore?

samanehsan: Oh, I see now! Previously the listener was given a path to a zip file with all of the dependencies, but the latest listener is getting all of the files from github so this makes sense

ambrosejcarr:
All comments responded to. Thanks, @samanehsan .

adapter_pipelines/ss2_single_sample/dependencies.json
samanehsan: Thanks for pointing that out @rexwangcc! I think it makes sense to include links to the dependency files in the config directly since that is where the links to the wdl and its input files are also defined.

Right now the listener is expecting a list of analysis wdls in the config file https://github.com/HumanCellAtlas/lira/blob/master/lira/api/notifications.py#L45 so I think we can remove this file @ambrosejcarr

adapter_pipelines/ss2_single_sample/adapter_example_static.json
```diff
@@ -0,0 +1,23 @@
...
+  "AdapterSmartSeq2SingleCell.star_ref_index": "gs://broad-dsde-mint-dev-teststorage/reference/GRCh38_Gencode/GRCh38_GencodeV27_Primary.tar",
+  "AdapterSmartSeq2SingleCell.hisat2_ref_index": "gs://broad-dsde-mint-dev-teststorage/reference/HISAT2/genome_snp_tran.tar.gz",
+  "AdapterSmartSeq2SingleCell.genome_ref_fasta": "gs://broad-dsde-mint-dev-teststorage/reference/GRCh38_Gencode/GRCh38.primary_assembly.genome.fa",
+  "AdapterSmartSeq2SingleCell.hisat2_ref_trans_index": "gs://broad-dsde-mint-dev-teststorage/reference/HISAT2/gencode_v27_trans_rsem.tar.gz",
+  "AdapterSmartSeq2SingleCell.rsem_ref_index": "gs://broad-dsde-mint-dev-teststorage/reference/rsem_ref/gencode_v27_primary.tar",
+  "AdapterSmartSeq2SingleCell.gene_ref_flat": "gs://broad-dsde-mint-dev-teststorage/reference/GRCh38_gencode.v27.refFlat.txt",
+  "AdapterSmartSeq2SingleCell.hisat2_ref_name": "genome_snp_tran",
+  "AdapterSmartSeq2SingleCell.stranded":"NONE",
+  "AdapterSmartSeq2SingleCell.reference_bundle": "bf51d668-3e14-4843-9bc7-5d676fdf0e01",
+  "AdapterSmartSeq2SingleCell.format_map": "gs://broad-dsde-mint-dev-teststorage/format_map_example.json",
```
rexwangcc: @ambrosejcarr I remembered @jishuxu has made a new format map file on the dev bucket (the format_map_example_SS2_latest.json`)for the new ss2, so that those files are now shown as unknown on the Ingest side, would you mind update it here? Or more easily, we rename that new file to this name (format_map_example.json).

rexwangcc: I have updated the file names in both staging and dev bucket.

ambrosejcarr: Hi Rex, confirming that based on your actions, we don't need to do any further modifications and this is good to merge?

adapter_pipelines/ss2_single_sample/dependencies.json
ambrosejcarr: Alright, if that's clear. Might support keeping it in master so things are clear, but agree that it's not needed here. :+1:

adapter_pipelines/ss2_single_sample/adapter_example_static.json
rexwangcc: @ambrosejcarr No, I renamed the filename in the bucket to match this line, we don't need to change anything here!

PR #13, opened 2018-01-16
Optimus wrapper updates

samanehsan:
Changes: 
- Change the example bundle to 10x v2 demo data 
- Remove input files for bundles that follow the old v3 metadata schema
- Remove ipython notebook from local testing


adapter_pipelines/Optimus/adapter_example_bundle_specific_v4.json
```diff
@@ -1,4 +1,4 @@
 {
-  "AdapterOptimus.bundle_uuid": "52a56d80-5d6b-4065-8931-2cd666951397",
-  "AdapterOptimus.bundle_version": "2017-12-14T222609.934808Z"
+  "AdapterOptimus.bundle_uuid": "9e9956b0-f731-4bc1-a000-2b73b5e735b3",
+  "AdapterOptimus.bundle_version": "2018-01-16T163044.805931Z"
```
rexwangcc: What do you think if we rename this file to `adapter_example_bundle_specific_v4_6_1.json`? This may be much clearer but require more changes everywhere though(such as config files of Lira)

adapter_pipelines/Optimus/adapter_example_bundle_specific_v3.json
```diff
@@ -1,4 +0,0 @@
-{
-  "AdapterOptimus.bundle_uuid": "c1677888-59fb-43bd-b3e3-05c274cdb11a",
-  "AdapterOptimus.bundle_version": "2017-09-21T173007.598036Z"
-}
```
rexwangcc: If we removed these v3 files for Optimus, I think it would be great we removed other v3 files for 10x and ss2, for a better consistency. What do you think?

samanehsan: That makes sense to me!

adapter_pipelines/Optimus/adapter_example_bundle_specific_v4.json
samanehsan: Every time the schema version changes there will be many files that would have to be renamed, which I would like to avoid. Maybe `"AdapterOptimus.schema_version": "4.6.1"` should go in this file instead since it's specific to the bundle. What do you think @rexwangcc?

rexwangcc: > Every time the schema version changes there will be many files that would have to be renamed, which I would like to avoid
I think this is really convincing to me. 

For the "AdapterOptimus.schema_version", there is already one in the static input json. If you're talking about adding an *indicator* field here, which just labels this bundle specific json's version for developers without any extra functionalities, I agree with you!

rexwangcc: But we need to give the field name carefully to prevent confusion, maybe something like: `"AdapterOptimus.data_bundle.schema_version"` is a little bit better than just `"AdapterOptimus.schema_version"`?

samanehsan: Oh, I was suggesting the schema_version field belongs in the bundle info sent from the data storage service, but that would require a change on their end! Do you think having the schema version defined in both files would be confusing?

rexwangcc: Oh, I see what you mean, I think having the same name(schema_version) defined in both files won't be confusing if they are pointing to the same thing and having exactly the same value. And I assume it will take a while before dss could add this field to their API. But this sounds like a good idea to me! What do you think? @dshiga

dshiga: I don't think we should have version numbers in any of these file names. Adding a "schema_version" field sounds okay to me.

samanehsan: Ok!

samanehsan:
I was just waiting for the CI build to pass, but it looks like Travis has a large backlog of builds that's
 causing delays 😞

PR #15, opened 2018-01-29
Turn on call-caching for all of the pipelines by default.

rexwangcc:
Turn on call-caching for all of the pipelines by default. This will speed up our integration test significantly.


rexwangcc:
Going to merge this PR to fix the integration test.

PR #20, opened 2018-02-22
Update paths to reference files to match their new locations (539)

dshiga:
Several reference files were moved as part of https://github.com/HumanCellAtlas/skylab/pull/84. This PR updates the static inputs json file to use the new paths.


samanehsan:
Just merged this now so that our mintegration tests will pass!

PR #22, opened 2018-03-02
Cw remove metadata schema checkers (575)

rexwangcc:
This PR removes the metadata schema file checker to avoid hitting the Github request limitation. Besides, it also:
- Updates submit.wdl to use latest pipeline-tools docker image.
- Removes deprecated `stage.py` from the pipeline-tools package.
- Updates outdated input files and wdls.


adapter_pipelines/submit.wdl
```diff
@@ -69,7 +69,7 @@ task create_submission {
   >>>
 
   runtime {
-    docker: "humancellatlas/pipeline-tools:0.1.5"
+    docker: "quay.io/humancellatlas/secondary-analysis-pipeline-tools:v0.1.9"
```
rexwangcc: I start to feel that putting adapter wdls and pipeline-tools package together in one repo makes us awkward now. 

Think about the below scenario: we find a bug in pipeline-tools python lib, say `v0.2.0`, which is referred in submit.wdl as `quay.io/humancellatlas/secondary-analysis-pipeline-tools:v0.2.0`. After fixing the bug we tag the repo with `v0.2.1` so we could have the auto-built docker image `v0.2.1`. We then turn to change the docker tag in submit.wdl from `v0.2.0` to `v0.2.1`. However, to utilize this new wdl file from lira, we need to either tag it with `v0.2.2` or move the previous tag to point to the new commit, both ways are not elegant.

Unless we could find a way to deal with this problem, we may need to either make wdls and python lib separate or manually build&push docker images.

samanehsan: That's true, unless we also update the docker tag in the submit.wdl at the same time as fixing the bug.

rexwangcc: Yep, but changing them at the same time, I guess, won't let us pass the Mintegration test?

PR #28, opened 2018-03-20
Add stub submit wdl

dshiga:
See https://elastc.com/c/4btDIPv3/615-stub-submission-wdl

It needs the same name as the regular submit wdl so that it will work with existing adapter wdls without changes. (Existing adapter wdls import a file named submit.wdl.)

### DIFF
--- a/adapter_pipelines/submit_stub/submit.wdl
+++ b/adapter_pipelines/submit_stub/submit.wdl
@@ -1,0 +1,27 @@
+task submit_stub {
+  command <<<
+    echo "This is a stub submission task for testing purposes only."
+  >>>
+  runtime {
+    docker: "python:2.7"
+  }
+}
+
+workflow submit {
+  Array[Object] inputs
+  Array[File] outputs
+  File format_map
+  String submit_url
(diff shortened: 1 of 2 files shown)


adapter_pipelines/submit_stub/submit.wdl
```diff
@@ -0,0 +1,27 @@
+task submit_stub {
+  command <<<
+    echo "This is a stub submission task for testing purposes only."
+  >>>
+  runtime {
+    docker: "python:2.7"
```
rexwangcc: If this stub `submit.wdl` doesn't do anything indeed, I would recommend using a smaller docker image here for minimizing the cost. [alpine](https://hub.docker.com/_/alpine/) could be a good candidate here.

dshiga: Tried alpine and also bash image and they're not working. Fortunately, this job will call cache every time since inputs never change, so image will only need to be pulled once.

rexwangcc: Glad to know that, thanks!

PR #31, opened 2018-03-23
Remove dss_url and submit_url from static inputs

samanehsan:
Remove `dss_url` and `submit_url` from static inputs, since they will vary based on the lira environment in https://github.com/HumanCellAtlas/lira/pull/67


samanehsan:
retest this please

PR #36, opened 2018-04-09
Retry getting metadata from Cromwell

samanehsan:
Changes in this PR: 
- Move embedded python code from get_metadata task into a python script 
- Use the retry decorator from the [tenacity](https://github.com/jd/tenacity) python library to retry the request to get metadata from Cromwell, with exponentially larger wait times in between requests. 
- Add unit tests for functions in the new script

---
- [x] Added or updated tests
- [x] Updated documentation

### DIFF
--- a/pipeline_tools/get_analysis_metadata.py
+++ b/pipeline_tools/get_analysis_metadata.py
@@ -9,8 +9,8 @@ def get_workflow_id(analysis_output_path):
     """
     Parse the analysis workflow id from one of its output paths, and write the id to a file so that it is available
     outside of the get_analysis task.
-    :param String analysis_output_path: Path to workflow output file
-    :return: String workflow_id
+    :param str analysis_output_path: Path to workflow output file.
+    :return: str workflow_id: Cromwell UUID of the workflow.
     """
     url = analysis_output_path
     hash_end = url.rfind("/call-")
@@ -24,8 +24,8 @@ def get_workflow_id(analysis_output_path):
 def get_auth(credentials_file=None):
     """
     Parse cromwell username and password from credentials file.
-    :param credentials_file: File containing cromwell authentication credentials
-    :return: HTTPBasicAuth object
+    :param str credentials_file: Path to the file containing cromwell authentication credentials.
+    :return: requests.auth.HTTPBasicAuth: Auth object to use for cromwell requests.
     """
     credentials_file = credentials_file or '/cromwell-metadata/cromwell_credentials.txt'
     with open(credentials_file) as f:
@@ -40,8 +40,8 @@ def get_metadata(runtime_environment, workflow_id):
     """
     Get metadata for analysis workflow from Cromwell and write it to a JSON file. Retry the request with exponentially
     increasing wait times if there is an error.
-    :param runtime_environment: The cromwell environment the workflow was run in
-    :param workflow_id: The analysis workflow id
+    :param str runtime_environment: The cromwell environment the workflow was run in.
+    :param str workflow_id: The analysis workflow id.
     :return:
     """
     print('Getting metadata for workflow {}'.format(workflow_id))


samanehsan:
retest this please

samanehsan:
please retest this

samanehsan:
please retest this

pipeline_tools/tests/test_get_analysis_metadata.py
```diff
@@ -0,0 +1,70 @@
...
+            context.status_code = 500
+            return {"status": "error", "message": "Internal Server Error"}
+
+        mock_auth.return_value = HTTPBasicAuth('user', 'password')
+        # Makes the test complete faster by limiting the number of retries
+        get_analysis_metadata.get_metadata.retry.stop = stop_after_attempt(3)
+
+        mock_request.get(self.cromwell_url, json=_request_callback)
+
+        with self.assertRaises(RetryError):
```
samanehsan: The `retry` decorator will raise a `RetryError` if it has to retry the request and fails, but there should be a way to have it raise the error that the request raises instead. Is there one that's preferred here?

samanehsan:
retest this please

samanehsan:
retest this please

pipeline_tools/tests/test_get_analysis_metadata.py
rexwangcc: I prefer the latter way actually. We'd better not let the tenacity to wrap the exceptions the original function throw. 

I tried the following code and it worked:
```python
@retry(reraise=True, stop=stop_after_attempt(3))
def raise_my_exception():
    raise ValueError("Fail")
raise_my_exception()
```

pipeline_tools/get_analysis_metadata.py
```diff
@@ -0,0 +1,48 @@
...
+        json.dump(response.json(), f, indent=2)
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--analysis_output_path', required=True)
+    parser.add_argument('--runtime_environment', required=True)
+    args = parser.parse_args()
+    workflow_id = get_workflow_id(args.analysis_output_path)
+    get_metadata(args.runtime_environment, workflow_id)
```
rexwangcc: I personally prefer to use something like `get_metadata.retry_with(wait=wait_exponential(multiplier=1, max=10), stop=stop_after_delay(20))(args.runtime_environment, workflow_id)` here explicitly so that we could control the retry parameters without going back to change the decorator parameters. But this is totally **OPTIONAL**

pipeline_tools/get_analysis_metadata.py
```diff
@@ -0,0 +1,48 @@
+import argparse
+import json
+import requests
+from requests.auth import HTTPBasicAuth
+from tenacity import retry, wait_exponential, stop_after_delay
+
+
+def get_workflow_id(analysis_output_path):
```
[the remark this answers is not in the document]
samanehsan: Ok! I left it as-is, but there is a ticket to fix this parsing!

pipeline_tools/tests/test_get_analysis_metadata.py
samanehsan: That sounds good to me!

pipeline_tools/get_analysis_metadata.py
samanehsan: Ok! I actually prefer the decorator pattern, but I think this would be useful if there are multiple requests we want to retry in the same function

samanehsan: Oh, or to pass in the retry parameters to the function which I think is what you meant originally! Good idea!

rexwangcc: Oh yeah, that is I originally meant! You read my mind LOL

pipeline_tools/get_analysis_metadata.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    Parse the analysis workflow id from one of its output paths, and write the id to a file so that it is available
+    outside of the get_analysis task.
+    :param String analysis_output_path: Path to workflow output file
+    :return: String workflow_id
+    """
```
rexwangcc: nit: 
```python
:param str analysis_output_path: Path to workflow output file
:return str workflow_id: Cromwell UUID of the workflow.
```
"The pain of having docstrings"  😄

pipeline_tools/get_analysis_metadata.py
```diff
@@ -0,0 +1,71 @@
...
+    with open('workflow_id.txt', 'w') as f:
+        f.write(workflow_id)
+    return workflow_id
+
+
+def get_auth(credentials_file=None):
+    """
+    Parse cromwell username and password from credentials file.
+    :param credentials_file: File containing cromwell authentication credentials
+    :return: HTTPBasicAuth object
```
rexwangcc: nit:
```python
:param str credentials_file: Path to the file containing cromwell authentication credentials.
:return requests.auth.HTTPBasicAuth: XXX
```

pipeline_tools/get_analysis_metadata.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@retry(reraise=True, wait=wait_exponential(multiplier=1, max=10), stop=stop_after_delay(20))
+def get_metadata(runtime_environment, workflow_id):
+    """
+    Get metadata for analysis workflow from Cromwell and write it to a JSON file. Retry the request with exponentially
+    increasing wait times if there is an error.
+    :param runtime_environment: The cromwell environment the workflow was run in
+    :param workflow_id: The analysis workflow id
+    :return:
```
rexwangcc: nit: 
```python
:param str runtime_environment: The cromwell environment the workflow was run in.
:param str workflow_id: The analysis workflow id.
```

PR #21, opened 2018-03-02
Update submit wdl to use cromwell-as-a-service

samanehsan:
If using cromwell-as-a-service, get the bearer token from the service account json key and send a request to caas to get the workflow metadata.

PR #38, opened 2018-04-13
Update submit wdl to use cromwell-as-a-service

samanehsan:
If using cromwell-as-a-service, get the bearer token from the service account json key and send a request to caas to get the workflow metadata.

Note: This is the same as https://github.com/HumanCellAtlas/pipeline-tools/pull/21, just under the same branch name as the corresponding Lira PR so that the integration test uses these branches together.

- [x] Added or updated tests
- [x] Updated documentation

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ WORKDIR /tools
 
 COPY . .
 
-RUN pip install .
+RUN pip install . --process-dependency-links
 
 # Install the latest hca-cli for submit.wdl
 RUN pip install hca==3.3.0
--- a/pipeline_tools/get_analysis_metadata.py
+++ b/pipeline_tools/get_analysis_metadata.py
@@ -3,6 +3,7 @@ import json
 import requests
 from requests.auth import HTTPBasicAuth
 from tenacity import retry, wait_exponential, stop_after_delay
+from cromwell_tools import cromwell_tools
 
 
 def get_workflow_id(analysis_output_path):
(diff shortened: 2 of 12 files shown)


samanehsan:
please retest this

samanehsan:
retest this please

pipeline_tools/get_analysis_metadata.py
```diff
@@ -36,17 +37,29 @@ def get_auth(credentials_file=None):
...
+        cromwell_url = 'https://cromwell.caas-dev.broadinstitute.org/api/workflows/v1'
+        json_credentials = caas_key_file or "/cromwell-metadata/caas_key.json"
+        headers = cromwell_tools.generate_auth_header_from_key_file(json_credentials)
+        auth = None
+    else:
+        cromwell_url = 'https://cromwell.mint-{}.broadinstitute.org/api/workflows/v1'.format(runtime_environment)
+        headers = None
+        auth = get_auth()
+
+    url = '{}/{}/metadata?expandSubWorkflows=true'.format(cromwell_url, workflow_id)
```
rexwangcc: Just FYI: recently I ran into an error when I was trying to use "expandSubWorkflows": https://github.com/broadinstitute/cromwell/issues/3209 This error should be fixed in newer Cromwell versions, but not in our current deployed Cromwell instances.

requirements.txt
```diff
@@ -1,2 +1,4 @@
 requests==2.18.4
-google-cloud-storage==1.8.0
\ No newline at end of file
+google-cloud-storage==1.8.0
+tenacity==4.10.0
+git+git://github.com/broadinstitute/cromwell-tools.git@2c9518ec869cea188dc5600dee148c4d9582eca1
```
rexwangcc: Could we update the hash code here with git tags after the other PR is merged in the Cromwell-tools?

samanehsan: Yes! This is just for testing until the cromwell-tools PR is merged in

samanehsan:
retest this please

samanehsan:
retest this please

samanehsan:
retest this please

samanehsan:
retest this please

Dockerfile
```diff
@@ -10,7 +10,7 @@ WORKDIR /tools
 
 COPY . .
 
-RUN pip install .
+RUN pip install . --process-dependency-links
```
rexwangcc: What does this `--process-dependency-links` do here?

rexwangcc: nvm, I just noticed this is used to process `setup.py` below

setup.py
```diff
@@ -25,5 +26,8 @@
               'confirm-submission=pipeline_tools.confirm_submission:main'
           ]
       },
+      dependency_links=[
+        'git+git://github.com/broadinstitute/cromwell-tools.git@2c9518ec869cea188dc5600dee148c4d9582eca1#egg=cromwell-tools-1.0.1'
```
rexwangcc: This is good for now but looks really hacky unfortunately.

samanehsan:
>Figure out how to remove the duplications in requirements.txt and setup.py, it's better we choose one and stick with it.

Looking at https://packaging.python.org/discussions/install-requires-vs-requirements/, our `setup.py` file should list out all of the packages it minimally needs in order to run, so that these dependencies get installed when the package is installed. So I think it's okay as it currently is, it just needs less strict versions as you mentioned. Our `requirements.txt` file is not being used at the moment though which was definitely confusing to me!

PR #39, opened 2018-04-19
Use offical blessed package to control the version(700)

rexwangcc:
This PR follows the [instruction](https://pypi.org/project/setuptools_scm/) to manage the version of Pipeline-tools. This will be not only necessary for Lira's `/version` endpoint but also important to maintain the versions of all of our dependencies.

So in other applications, we can now get the version by:
```python
import pipeline_tools

pipeline_tools.__version__
```
And we don't need to update the `setup.py` manually anymore, it will automatically grab the right git tag.

- [ ] Added or updated tests
- [ ] Updated documentation


rexwangcc:
Rebased and addressed the comments, will merge it.

PR #40, opened 2018-04-26
Retry with other boxes

samanehsan:
Add retries when making requests to the HCA data storage service and ingestion service, using the [tenacity](https://github.com/jd/tenacity) python library to add exponential wait times and timeouts. Also simplify the retry logic for getting the staging urn and checking if the submission status is valid.

Note: I added tests for the retry behavior and noticed some calls to ingest do not currently have unit tests. Creating these tests seem to overlap with the ticket on recording responses from calls to other boxes for our border tests, so I've left them as test stubs for now.

- [x] Added or updated tests
- [x] Updated documentation


samanehsan:
retest this please

samanehsan:
retest this please

samanehsan:
retest this please

pipeline_tools/confirm_submission.py
```diff
@@ -2,44 +2,66 @@
...
+
+
+@retry(reraise=True, wait=wait_exponential(multiplier=1, max=RETRY_SECONDS), stop=stop_after_delay(TIMEOUT_SECONDS))
 def confirm(envelope_url):
     print('Confirming submission')
     headers = {
         'Content-type': 'application/json'
     }
     response = requests.put('{}/submissionEvent'.format(envelope_url), headers=headers)
+    response.raise_for_status()
```
rexwangcc: dumb question: would the errors threw here break the retry?

pipeline_tools/create_envelope.py
```diff
@@ -4,56 +4,94 @@
...
+                                             stop=stop_after_delay(timeout_seconds))(analyses_url, auth_headers, analysis_json_contents)
+
+    # 4. Add input bundles
+    input_bundles_url = get_entity_url(analysis_js, 'add-input-bundles')
+    add_input_bundles.retry_with(wait=wait_exponential(retry_seconds),
+                                 stop=stop_after_delay(timeout_seconds))(input_bundles_url, auth_headers, analysis_json_contents)
+
+    # 5. Add file references
+    file_refs_url = get_entity_url(analysis_js, 'add-file-reference')
+    print('Adding file references at {0}'.format(file_refs_url))
```
rexwangcc: nit: These prints are moved from legacy code, but it would be really great if we could use logging module here! To make loggers output to both logs and stdouts, there is an example of using 2 handlers: https://stackoverflow.com/questions/14058453/making-python-loggers-output-all-messages-to-stdout-in-addition-to-log 

If there is too much to do here, we can make it separate from this PR tho.

pipeline_tools/tests/data/response.json
```diff
@@ -1,47 +1,47 @@
...
+    },
+    "user": {
+      "href": "http://api.ingest.dev.data.humancellatlas.org/user"
+    },
+    "profile": {
+      "href": "http://api.ingest.dev.data.humancellatlas.org/profile"
     }
   }
-}
+}
```
rexwangcc: nit: new line

pipeline_tools/tests/data/response.json
```diff
@@ -1,47 +1,47 @@
...
-      "title" : "Access or create, within a submission envelope, a new assay"
-    },
-    "submissionEnvelopes" : {
-      "href" : "http://api.ingest.dev.data.humancellatlas.org/submissionEnvelopes{?page,size,sort}",
-      "templated" : true,
-      "title" : "Access or create new submission envelopes"
-    },
-    "profile" : {
-      "href" : "http://api.ingest.dev.data.humancellatlas.org/profile"
+  "_links": {
```
rexwangcc: updating this file for v5 is so cooool!

pipeline_tools/tests/test_confirm_submission.py
```diff
@@ -0,0 +1,62 @@
...
+    def test_wait_for_valid_status(self, mock_request):
+        envelope_url = 'http://api.ingest.dev.data.humancellatlas.org/submissionEnvelopes/abcde'
+
+        def _request_callback(request, context):
+            context.status_code = 200
+            return {'submissionState': 'Valid'}
+
+        mock_request.get(envelope_url, json=_request_callback)
+        status = confirm_submission.wait_for_valid_status(envelope_url)
+        self.assertEqual(status, 'Valid')
```
rexwangcc: it would be better to check if the `call_count` is 1 here

pipeline_tools/tests/test_confirm_submission.py
```diff
@@ -0,0 +1,62 @@
...
+
+        def _request_callback(request, context):
+            context.status_code = 200
+            return {'submissionState': 'Validating'}
+
+        mock_request.get(envelope_url, json=_request_callback)
+        with self.assertRaises(RetryError):
+            # Make the test complete faster by limiting the number of retries
+            confirm_submission.wait_for_valid_status.retry_with(stop=stop_after_attempt(3))(envelope_url)
+        self.assertNotEqual(mock_request.call_count, 1)
```
rexwangcc: Should it also check `call_count == 3` as well?

pipeline_tools/confirm_submission.py
samanehsan: Not a dumb question at all! Throwing an error here will cause the function to retry until there is a successful response and timeout after the specified `TIMEOUT_SECONDS`. If it times out, it will raise the HTTP error from the request.

pipeline_tools/tests/test_confirm_submission.py
samanehsan: That makes sense!

pipeline_tools/create_envelope.py
samanehsan: Is it okay to make this change in a separate PR? I agree logging would be a great improvement, but this PR is already pretty big!

rexwangcc: Sure! I agree to separate it in a separate PR, and also, it is a "Should-have" instead of a "Mut have" item.

samanehsan:
retest this please

samanehsan:
Thanks for your comments @rexwangcc! I also made one more change in my last commit if you could take another look. Sorry I forgot to add it in earlier!

pipeline_tools/create_envelope.py
```diff
@@ -4,79 +4,94 @@
...
-    print('Adding file references at {0}'.format(file_refs_url))
-    output_files = get_output_files(analysis_json_contents, schema_version)
-    for file_ref in output_files:
-        print('Adding file: {}'.format(file_ref['fileName']))
-        response = requests.put(file_refs_url, headers=auth_headers, data=json.dumps(file_ref))
-        check_status(response.status_code, response.text)
-
-
-def check_status(status, response_text, expected='2xx'):
-    """Check that status is in range 200-299 or the specified range, if given.
```
rexwangcc: I'm not sure if `raise_for_status` can fully replace this function? It seems this function is not only checking if the status code is 2xx, but also compare the status code with the `expected` parameter. 

The question is do we really have a use case that requires us inputting the `expected` parameter? If not, we can replace this with `raise_for_status`!

rexwangcc: Looks good to me! Thanks for this change!

PR #43, opened 2018-04-27
Record HTTP requests and responses

dshiga:
Adds the ability to record HTTP requests and responses by writing them to files.

Details:
- Added an HttpRequests class containing wrapper functions for requests.get, put, and post. The wrapper functions perform the recording. Also centralized tenacity retry code in HttpRequests to make it easier to modify timeouts, etc. in one place.
- WDL tasks are now populating some environment variables to  configure retries and turn recording on or off.
- 4xx errors are not retried any more -- better to fail quickly if we're making bad requests
- Added tests
- Changed most doc strings to Google style

I'd like to modify pipeline-tools in future to make it all one big CLP, at which point we might stop configuring things via environment variables and instead configure via optional params passed into the main CLP. But that should be its own ticket.

See: https://elastc.com/c/DfIYLZp*/672-record-requests-and-responses-in-adapter-wdls

- [x] Added or updated tests
- [x] Updated documentation


dshiga:
retest this please

dshiga:
retest this please

dshiga:
retest this please

dshiga:
retest this please

dshiga:
retest this please

dshiga:
retest this please

dshiga:
retest this please

dshiga:
retest this please

dshiga:
retest this please

dshiga:
retest this please

adapter_pipelines/submit.wdl
```diff
@@ -2,20 +2,32 @@
...
+  Int? timeout_seconds
   Boolean use_caas
+  Boolean record_http
 
   command <<<
+    export RECORD_HTTP_REQUESTS="${record_http}"
+    export RETRY_TIMEOUT="${timeout_seconds}"
+    export RETRY_MULTIPLIER="${retry_multiplier}"
+    export RETRY_MAX_INTERVAL="${retry_max_interval}"
+    touch request_000.txt && touch response_000.txt
```
rexwangcc: I'm not sure about why it has to touch the files here?

pipeline_tools/confirm_submission.py
```diff
@@ -2,67 +2,81 @@
...
+        http_requests: HttpRequests object
 
+    Returns:
+        Status of the submission ("Valid", "Validating", etc.)
+
+    Raises:
+        requests.HTTPError if response status indicates an error
+    """
+    def log_before(envelope_url):
+        print('Getting status for {}'.format(envelope_url))
```
rexwangcc: I've seen this `log_before` function elsewhere in this PR, how about taking it out of this nested functions and making it more general? like: 
```python
    def log_before(current_stage, envelope_url):
        print('Getting'{0}' for {1}'.format(current_stage, envelope_url))
```

pipeline_tools/confirm_submission.py
```diff
@@ -2,67 +2,81 @@
...
+    Args:
+        envelope_url: a string representing the url for the envelope
+        http_requests: HttpRequests object
+
+    Returns:
+        The text of the response
+
+    Raises:
+        requests.HTTPError if the response status indicates an error
+    """
```
rexwangcc: nit: 
```
Args:
    envelope_url (str): a string representing the url for the envelope
    http_requests (HttpRequests): HttpRequests object

Returns:
    unicode|str: The text of the response, unicode for Python2.x and str for Python3

Raises:
    requests.HTTPError: If response status indicates an error
```

pipeline_tools/create_analysis_json.py
```diff
@@ -2,22 +2,42 @@
...
+    Returns:
+        Dictionary representing the analysis json file to be submitted
+    """
     print('Creating analysis.json for {}'.format(analysis_id))
 
     with open(metadata_file) as f:
         metadata = json.load(f)
-        start, end = get_start_end(metadata)
+        start = metadata['start']
+        end = metadata['end']
```
rexwangcc: nit: collapse two lines into one: `start, end = metadata['start'], end = metadata['end']`

pipeline_tools/create_analysis_json.py
```diff
@@ -2,22 +2,42 @@
...
+        print(start, end)
         tasks = get_tasks(metadata)
 
     inputs = create_inputs(inputs_file)
     outputs = create_outputs(outputs_file, format_map, schema_version)
 
-    input_bundles = get_input_bundles(input_bundles_string)
+    input_bundles = input_bundles_string.split(',')
+    print(input_bundles)
```
rexwangcc: Would be better to have some information around this:, e.g. `print('Input Bundles: {}.format(input_bundles)')`

pipeline_tools/http_requests.py
```diff
@@ -0,0 +1,272 @@
...
+        should_record: A boolean indicating whether to record requests and responses.
+        record_dir: A string indicating the directory where requests and responses
+            should be written.
+    """
+
+    def __init__(self, write_dummy_files=True):
+        """Initializes should_record and record_dir from environment variables.
+        Also writes dummy request_000.txt and response_000.txt files needed to prevent
+        an error when using this class in WDL and glob is used to find the outputs.
+        """
```
rexwangcc: I see why you `touch` those two files in the WDL, nvm my previous comments!

pipeline_tools/http_requests.py
```diff
@@ -0,0 +1,272 @@
...
+        self.retry_timeout = self._get_param(RETRY_TIMEOUT, '600', int)
+        self.retry_max_tries = self._get_param(RETRY_MAX_TRIES, '10000', int) 
+        self.retry_multiplier = self._get_param(RETRY_MULTIPLIER, '1', float)
+        self.retry_max_interval = self._get_param(RETRY_MAX_INTERVAL, '10', float)
+
+    def _get_param(self, name, default, convert_fn=None):
+        param_str = os.environ.get(name, default)
+        # handle case where environment var is set to empty string
+        if not param_str:
+            param_str = default
```
rexwangcc: nit: 1-liner: `param_str = param_str or default`

pipeline_tools/http_requests.py
```diff
@@ -0,0 +1,272 @@
...
+
+        if convert_fn is not None:
+            return convert_fn(param_str)
+        else:
+            return param_str
```
rexwangcc: nit: 1-liner: `return param_str if not convert_fn else convert_fn(param_str)`

pipeline_tools/http_requests.py
```diff
@@ -0,0 +1,272 @@
...
+        return self._http_request_with_retry(*args, **kwargs)
+
+    def _http_request_with_retry(self, *args, **kwargs):
+        def _is_retryable_status_code(error):
+            if not isinstance(error, requests.HTTPError):
+                return False
+            return not (400 <= error.response.status_code <= 499)
+
+        if 'retry' in kwargs:
+            retry = kwargs['retry'] | retry_if_exception(_is_retryable_status_code)
```
rexwangcc: unless it is doing bit-wise OR here, I would prefer `or` over `pipe |`

pipeline_tools/http_requests.py
```diff
@@ -0,0 +1,272 @@
...
+        Returns:
+            A string representing the next suffix, one higher than the highest one present
+            in the given list.
+        """
+        if len(files) == 0:
+            return '001'
+        sorted_files = sorted(files, reverse=True)
+        match = re.search('_([\d]+)\.txt', sorted_files[0])
+        index = int(match.group(1)) + 1
+        # Zero pad number so it has 2 digits
```
rexwangcc: 3 digits?

pipeline_tools/http_requests.py
rexwangcc: If I understand it correctly, does the `kwargs['retry']` here used for keeping the ability to tweak/fine-tune retry strategy for each individual functions for us? If this is the case, strong 👍 for it. My concern about putting retry parameters into env variables is that we lose the ability to tweak the retry strategy (e.g. retry timeout seconds) for various individual functions.

rexwangcc:
I dropped a few comments. Most of the Google-styled docstrings are kinda not well-formatted, so Sphinx might not be able to build and render the parameter types properly on readthedocs.io, (e.g. `envelope_url: String of ubmission envelope url` => `envelope_url (str): Submission envelope url`) But since this is already a great and surprisingly large PR, we could separate the work about docstrings style to a separate PR. Generally, it looks GOOD to me! Thanks!

samanehsan:
It looks like we'll also need to update the `adapter.wdl` and input json files for the optimus workflow, but that could go in a separate PR

pipeline_tools/confirm_submission.py
```diff
@@ -2,67 +2,81 @@
...
-
-
-def get_envelope_json(envelope_url):
-    response = requests.get(envelope_url)
-    envelope_js = response.json()
-    check_status(response.status_code, response.text)
-    return envelope_js
+    response = http_requests.put('{}/submissionEvent'.format(envelope_url), headers=headers)
+    text = response.text
+    print(text)
```
samanehsan: Do we still need to print out the response text to stdout now that we are recording the responses in a separate file?

pipeline_tools/http_requests.py
samanehsan: The `else` is not necessary, you could just `return param_str`

pipeline_tools/confirm_submission.py
```diff
@@ -2,67 +2,81 @@
...
+        print('Getting status for {}'.format(envelope_url))
+
+    def status_is_invalid(response):
+        envelope_js = response.json()
+        status = envelope_js.get('submissionState')
+        print('submissionState: {}'.format(status))
+        return status != 'Valid'
+
+    response = (
+        http_requests
```
samanehsan: Will this write the response.txt for each retried request? In this case, we can expect several retries until the status is `Valid`, so I'm wondering how many files that would create.

pipeline_tools/tests/test_confirm_submission.py
```diff
@@ -2,23 +2,27 @@
...
             return {'submissionState': 'Valid'}
 
         mock_request.get(envelope_url, json=_request_callback)
-        status = confirm_submission.wait_for_valid_status(envelope_url)
-        self.assertEqual(status, 'Valid')
-        self.assertEqual(mock_request.call_count, 1)
+
+        with HttpRequestsManager():
+            result = confirm_submission.wait_for_valid_status(envelope_url, HttpRequests())
+            self.assertEqual(result, True)
```
samanehsan: I think these assert statements should be outside of the `with` block

pipeline_tools/tests/test_get_analysis_metadata.py
```diff
@@ -65,17 +68,11 @@ def _request_callback(request, context):
...
-
-        with self.assertRaises(requests.HTTPError):
-            get_analysis_metadata.get_metadata(self.runtime_environment, self.workflow_id, use_caas=False)
-            self.assertNotEqual(mock_request.call_count, 1)
-
-        # Reset decorator default
-        get_analysis_metadata.get_metadata.retry.stop = stop_after_delay(20)
+        with self.assertRaises(requests.HTTPError), HttpRequestsManager():
+            get_analysis_metadata.get_metadata(self.runtime_environment, self.workflow_id, HttpRequests(), use_caas=False)
+            self.assertEqual(mock_request.call_count, 3)
```
samanehsan: This assert statement should be outside of the `with self.assertRaises` block so that the test can fail if it raises an error.

pipeline_tools/http_requests.py
samanehsan: How about including the parameters in the class constructor to allow optionally overriding the environment variables?

pipeline_tools/tests/test_confirm_submission.py
samanehsan: Oh actually I'm thinking about how assert statements inside of a `with self.assertRaises(requests.HTTPError)` block would not raise errors in the test. It may not matter in this case with the `HttpRequestsManager`

pipeline_tools/http_requests.py
```diff
@@ -0,0 +1,272 @@
...
+
+        Raises:
+            ValueError if response status code indicates an error.
+        """
+        kwargs['method'] = 'post'
+        return self._http_request_with_retry(*args, **kwargs)
+
+    def _http_request_with_retry(self, *args, **kwargs):
+        def _is_retryable_status_code(error):
+            if not isinstance(error, requests.HTTPError):
```
samanehsan: Could we simplify this to `return isinstance(error, requests.HTTPError) and not (400 <= error.response.status_code <= 499)`? I find checks like this easier to understand when there are no double-negatives!

samanehsan:
Thanks for making these changes @dshiga! I left a few comments/questions in the PR ☀️

pipeline_tools/confirm_submission.py
dshiga: It's a good thought but I'm not sure that will generalize well beyond these two cases and the code saved is very minimal. I started writing a more generalizable function like
```python
def log_before(format_str, *args):
    print(format_str.format(*args))
```

but didn't go through with it because although it reduces duplication it makes things more verbose in some ways, so it didn't seem like a win overall.

pipeline_tools/confirm_submission.py
dshiga: I think it's still worth doing since the recording is optional and might not always be turned on.

pipeline_tools/confirm_submission.py
dshiga: It could be a lot of files depending on the retry parameters used. I think we should 1) make sure we set the retry parameters wisely so we don't retry an inordinate number of times, which wouldn't serve much purpose 2) only turn recording on when we really need debugging info, such as for scale testing.

pipeline_tools/tests/test_get_analysis_metadata.py
dshiga: good catch, thanks

pipeline_tools/http_requests.py
dshiga: `kwargs['retry']` lets you override the way we decide whether to retry. By default, we retry only if there is a 5xx error, but we use the `kwargs['retry']` argument in a couple of places to let us retry based on the content of the returned json.

That's a good suggestion @samanehsan, but let's wait to add that ability until we really want to use it somewhere.

I'm hoping that we can switch from using environment variables to command line parameters in a future refactoring that was too big to undertake in this already big PR, so hopefully this situation is temporary anyway.

dshiga:
retest this please

pipeline_tools/confirm_submission.py
rexwangcc: That makes sense!

dshiga:
Addressed comments, please take another look.

pipeline_tools/http_requests.py
```diff
@@ -0,0 +1,285 @@
...
+        Raises:
+            requests.HTTPError: if 5xx error occurs
+            tenacity.RetryError: if retry limit is reached and condition specified by
+            retry kwarg is still not met
+        """
+        kwargs['method'] = 'post'
+        return self._http_request_with_retry(*args, **kwargs)
+
+    def _http_request_with_retry(self, *args, **kwargs):
+        def _is_retryable_status_code(error):
```
samanehsan: We'll also want to retry on connection timeouts here in addition to http errors

dshiga:
retest this please

dshiga:
retest this please

PR #48, opened 2018-05-30
Retry on readtimeout 782

rexwangcc:
Add retry logic for ReadTimeout exception to submission step, so that workflows will retry if Ingest is temporarily unavailable: https://elastc.com/c/ZH5icvdf/782-retry-on-readtimeout-exceptions. I'm not sure if I have added the retry logic in the proper way tho.

Besides, there are also some style fixes in the PR.

- [x] Added or updated tests
- [x] Updated documentation

### DIFF
--- a/pipeline_tools/confirm_submission.py
+++ b/pipeline_tools/confirm_submission.py
@@ -1,7 +1,8 @@
 #!/usr/bin/env python
 
 import argparse
-from tenacity import retry_if_result, RetryError
+import requests
+from tenacity import retry_if_result, RetryError, retry_if_exception_type
 from pipeline_tools.http_requests import HttpRequests
 
 
@@ -35,7 +36,7 @@ def wait_for_valid_status(envelope_url, http_requests):
         .get(
             envelope_url,
             before=log_before(envelope_url),
-            retry=retry_if_result(status_is_invalid)
+            retry=(retry_if_result(status_is_invalid) | retry_if_exception_type(requests.exceptions.ReadTimeout))
         )
     )
     return True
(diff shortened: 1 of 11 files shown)


pipeline_tools/confirm_submission.py
```diff
@@ -59,7 +59,10 @@ def confirm(envelope_url, http_requests):
     headers = {
         'Content-type': 'application/json'
     }
-    response = http_requests.put('{}/submissionEvent'.format(envelope_url), headers=headers)
+    response = http_requests.put('{}/submissionEvent'.format(envelope_url),
+                                 headers=headers,
+                                 retry=retry_if_exception_type(requests.exceptions.ReadTimeout)
```
samanehsan: In these requests, we still want to retry if the exception type is `requests.HTTPError` or `requests.ConnectionError` as well. Right now this overrides that default: https://github.com/HumanCellAtlas/pipeline-tools/blob/master/pipeline_tools/http_requests.py#L146

How about including `requests.exceptions.ReadTimeout` as a retryable status code here? https://github.com/HumanCellAtlas/pipeline-tools/blob/master/pipeline_tools/http_requests.py#L143

pipeline_tools/confirm_submission.py
```diff
@@ -36,7 +36,7 @@ def status_is_invalid(response):
         .get(
             envelope_url,
             before=log_before(envelope_url),
-            retry=retry_if_result(status_is_invalid)
+            retry=(retry_if_result(status_is_invalid) | retry_if_exception_type(requests.exceptions.ReadTimeout))
```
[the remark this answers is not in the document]
samanehsan: Nevermind! This will not override the default retry parameter

pipeline_tools/confirm_submission.py
samanehsan: As discussed in person, this actually does not override the default retry parameter! But it would still be simpler to include `requests.exceptions.ReadTimeout` as a default retryable error so that it does not have to be passed into each request.

PR #50, opened 2018-06-05
[WIP] Update analysis JSON to v6 metadata schema

samanehsan:
The type of the analysis json file changed from "process" to "protocol" in the HCA v6 metadata schema, mainly affecting the schema urls.


samanehsan:
Replaced by #53

PR #52, opened 2018-06-08
Fix the order of stage and confirm tasks.

rexwangcc:
Previously I separated the confirm_submission from the stage_files task, but since they are not depending on each other, Cromwell will try to execute them at the same time, which causes some of the tests fail because of the long retry time. This PR fixes this problem by forcefully letting the 2 tasks depend on each other. 

- [ ] Added or updated tests
- [ ] Updated documentation

### DIFF
--- a/adapter_pipelines/submit.wdl
+++ b/adapter_pipelines/submit.wdl
@@ -150,11 +150,15 @@ task stage_files {
   output {
     Array[File] http_requests = glob("request_*.txt")
     Array[File] http_responses = glob("response_*.txt")
+    # Make this task output a placeholder string here for confirm_submission so that they can be executed in order
+    String ready_to_be_confirmed = "true"
   }
 }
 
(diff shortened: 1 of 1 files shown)


adapter_pipelines/submit.wdl
```diff
@@ -150,11 +150,15 @@ task stage_files {
   output {
     Array[File] http_requests = glob("request_*.txt")
     Array[File] http_responses = glob("response_*.txt")
+    # Make this task output a placeholder string here for confirm_submission so that they can be executed in order
+    String ready_to_be_confirmed = "true"
```
samanehsan: I think it makes sense to use Cromwell's `Boolean` type here instead since the value is `true`

rexwangcc:
@samanehsan Addressed the comments, could you have another look when you have time? Thanks!

PR #53, opened 2018-06-08
Update to v6 metadata schema

samanehsan:
- Update getting the sample id from `links_json`
- Renamed `uuid` and `version` to `bundle_uuid` and `bundle_version` for clarity in `input_utils.py`
- The type of the analysis JSON file changed from "process" to "protocol"
- Schemas are independently versioned, so we need to explicitly specify the schema versions we're using. Right now there are only two that are required when creating the analysis JSON (all unnecessary ones were removed):
    - analysis_protocol file: https://schema.humancellatlas.org/type/protocol/analysis/{schema_version}/analysis_protocol
    - output file: https://schema.humancellatlas.org/type/file/{analysis_file_version}/analysis_file

- [x] Added or updated tests
- [ ] Updated documentation

### DIFF
--- a/adapter_pipelines/ss2_single_sample/adapter_example_static_demo.json
+++ b/adapter_pipelines/ss2_single_sample/adapter_example_static_demo.json
@@ -9,6 +9,7 @@
   "AdapterSmartSeq2SingleCell.format_map": "gs://hca-dcp-mint-test-data/adapters/file_format_map.json",
   "AdapterSmartSeq2SingleCell.method": "SmartSeq2SingleCell",
   "AdapterSmartSeq2SingleCell.schema_version": "7.0.0",
+  "AdapterSmartSeq2SingleCell.analysis_file_version": "5.2.1",
   "AdapterSmartSeq2SingleCell.run_type": "run",
   "AdapterSmartSeq2SingleCell.retry_seconds": 10,
   "AdapterSmartSeq2SingleCell.timeout_seconds": 1200
--- a/adapter_pipelines/ss2_single_sample/adapter_example_static_demo.json
+++ b/adapter_pipelines/ss2_single_sample/adapter_example_static_demo.json
@@ -8,7 +8,7 @@
   "AdapterSmartSeq2SingleCell.reference_bundle": "bf51d668-3e14-4843-9bc7-5d676fdf0e01",
   "AdapterSmartSeq2SingleCell.format_map": "gs://hca-dcp-mint-test-data/adapters/file_format_map.json",
   "AdapterSmartSeq2SingleCell.method": "SmartSeq2SingleCell",
-  "AdapterSmartSeq2SingleCell.schema_version": "5.1.0",
+  "AdapterSmartSeq2SingleCell.schema_version": "7.0.0",
   "AdapterSmartSeq2SingleCell.run_type": "run",
   "AdapterSmartSeq2SingleCell.retry_seconds": 10,
   "AdapterSmartSeq2SingleCell.timeout_seconds": 1200
--- a/adapter_pipelines/submit.wdl
+++ b/adapter_pipelines/submit.wdl
@@ -93,7 +93,7 @@ task create_submission {
   >>>
 
   runtime {
-    docker: "quay.io/humancellatlas/secondary-analysis-pipeline-tools:se-test-v6-changes"
+    docker: "quay.io/humancellatlas/secondary-analysis-pipeline-tools:v0.21.0"
   }
   output {
     File analysis_json = "analysis.json"
@@ -147,7 +147,7 @@ task stage_files {
   >>>
 
   runtime {
-    docker: "quay.io/humancellatlas/secondary-analysis-pipeline-tools:se-test-v6-changes"
+    docker: "quay.io/humancellatlas/secondary-analysis-pipeline-tools:v0.21.0"
   }
   output {
     Array[File] http_requests = glob("request_*.txt")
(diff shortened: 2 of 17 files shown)


samanehsan:
The mintegration test is failing because it's using a v5 bundle. The PR to update the bundle is here: https://github.com/HumanCellAtlas/secondary-analysis/pull/44

dshiga:
I think we've been using the wrong sample id all along.

Probably the sample id we really want is at: biomaterials_json.biomaterials.content.biomaterial_core.biomaterial_id
(there's more than one in each biomaterials.json file but we want the specimen_from_organism one)

For example it's value is "Q4_DEMO-sample_SAMN02797092" in this bundle:
https://dss.integration.data.humancellatlas.org/v1/bundles/f4b1199a-5c72-4c20-ae42-54862b361a72/?replica=aws

Here is the biomaterials json for that bundle:
https://dss.integration.data.humancellatlas.org/v1/files/8f97f492-956c-4e5c-b6f3-c32ef0bcab9c/?replica=aws

See [comment in slack](https://humancellatlas.slack.com/archives/C49TW3RV5/p1528845456000202).

samanehsan:
@dshiga, that's a good point! I believe right now we are using the biomaterial id of the cell suspension as the sample id. If we want to be able to track sample progress with the "sample id" label, it's probably better to still use the `uuid` instead of the user-given sample name.

adapter_pipelines/submit.wdl
```diff
@@ -211,6 +213,7 @@ workflow submit {
   Boolean use_caas
   # By default, don't record http requests
   Boolean record_http = false
+  String analysis_file_version = "5.2.1"
```
rexwangcc: nit: I'm not sure why it's necessary to put a default value here, since submit wdl should never run independently, and each time the metadata version gets updated, we have to change both here and the static input JSON for adapter wdl, even this variable is always overridden by the other value.

adapter_pipelines/ss2_single_sample/adapter_example_static_demo.json
```diff
@@ -8,7 +8,8 @@
   "AdapterSmartSeq2SingleCell.reference_bundle": "bf51d668-3e14-4843-9bc7-5d676fdf0e01",
   "AdapterSmartSeq2SingleCell.format_map": "gs://hca-dcp-mint-test-data/adapters/file_format_map.json",
   "AdapterSmartSeq2SingleCell.method": "SmartSeq2SingleCell",
-  "AdapterSmartSeq2SingleCell.schema_version": "5.1.0",
+  "AdapterSmartSeq2SingleCell.schema_version": "7.0.0",
+  "AdapterSmartSeq2SingleCell.analysis_file_version": "5.2.1",
```
rexwangcc: This I think is the best solution for us at this time. 

However, this is still painful for us. Per offline discussion, in case that major metadata schema version updates break our prod pipelines, we pin to specific versions here, so we only have to go back and update the code here when we are notified to do it, this requires other boxes (e.g. Purple) have the same behavior as we do (Pin to a specific version instead of always use the `latest` metadata version) and only update to new major versions coordinately. 

Ideally, as a consumer of the metadata, we should be able to use an abstract interface (API/CLI) instead of having to hardcode the parsing on our side and update frequently, as others are already discussing in the channels.

pipeline_tools/input_utils.py
```diff
@@ -129,15 +146,17 @@ def get_smart_seq_2_fastq_names(metadata, version):
     Raises:
         NotImplementedError: if metadata version is unsupported
     """
-    if version.startswith('5.'):
-        return _get_smart_seq_2_fastq_names_v5(metadata)
-    elif version.startswith('4.'):
+    version_prefix = int(version.split('.', 1)[0])
+    print(version_prefix)
```
rexwangcc: nit: remove this print, or make it more informative :)

samanehsan: Thanks! This was for debugging :)

pipeline_tools/input_utils.py
```diff
@@ -2,25 +2,46 @@
...
     Raises:
         NotImplementedError: if version is unsupported
     """
-    if version.startswith('4.'):
+    if schema_version.startswith('4.'):
         return _get_sample_id_v4(metadata)
-    elif version.startswith('5.'):
+    elif schema_version.startswith('5.'):
         return _get_sample_id_v5(metadata)
+    elif schema_version.startswith('6.'):
```
rexwangcc: Can you clarify here about how it could handle those verions startwith `7.`? Since I feel like `_get_smart_seq_2_fastq_names_v5_or_higher()` will possibly return `v7.0.0`?

adapter_pipelines/ss2_single_sample/adapter_example_static_demo.json
samanehsan: At the moment, ingest will accept submissions that follow older schema versions so it should accept this format even if they update this again. But they may change that behavior in the future and only accept and validate against the latest schema.

adapter_pipelines/submit.wdl
samanehsan: I added this default here because I was too lazy to update the Optimus and 10x adapter wdls in this PR to pass in the `analysis_file_version`, but I can add those in and remove this default so that's one less place to change the version

rexwangcc:
Reminder: changing the docker tags before merging :octocat: @samanehsan

samanehsan:
retest this please

samanehsan:
> regarding the 10x adapter wdl, should we remove it from the code base?

@rexwangcc if I remember correctly we decided we can remove the adapter wdl but leave the 10x wdl in skylab. Does that sound right @dshiga?

samanehsan:
Also the integration test was failing previously because of the race condition in building the `pipeline-tools` image in quay! This will still work with v5 bundles.

PR #55, opened 2018-06-19
Fix parsing analysis workflow id

samanehsan:
Get the first subworkflow id from the analysis output path, which corresponds to the analysis workflow id, instead of always getting the last workflow id in the path. 

- [x] Added or updated tests
- [ ] Updated documentation


rexwangcc:
retest this please

samanehsan:
@rexwangcc it looks like the failure was due to an intermittent error in the test because I could not reproduce it locally, and the test passed when I restarted the build.

PR #56, opened 2018-06-26
Use v4.0.0 of hca cli (838)

dshiga:
Updates to latest hca cli (v4.0.0) to deal with new upload uri format.

Also parameterized pipeline tools docker version so that in future we only need to change it in one place.

See: https://elastc.com/c/Z1m5xWfM/838-support-iam-not-from-upload-service

- [ ] Added or updated tests
- [x] Updated documentation

### DIFF
--- a/pipeline_tools/http_requests.py
+++ b/pipeline_tools/http_requests.py
@@ -234,12 +234,16 @@ class HttpRequests(object):
 
     @staticmethod
     def _get_method(http_method):
+        session = requests.Session()
+        # The default max_redirects is 30. We need to boost it because DSS sometimes redirects us more than 30 times,
+        # which causes an error if we don't allow for more.
+        session.max_redirects = 100
         if http_method == 'get':
-            fn = requests.get
+            fn = session.get
         elif http_method == 'put':
-            fn = requests.put
+            fn = session.put
         elif http_method == 'post':
-            fn = requests.post
+            fn = session.post
         else:
             raise NotImplementedError('Only get, put, and post are implemented.')
         return fn
--- a/adapter_pipelines/Optimus/adapter.wdl
+++ b/adapter_pipelines/Optimus/adapter.wdl
@@ -8,6 +8,7 @@ task GetInputs {
   String dss_url
   Int retry_seconds
   Int timeout_seconds
+  String pipeline_tools_version
 
   command <<<
     # Force the binary layer of the stdout and stderr streams to be unbuffered.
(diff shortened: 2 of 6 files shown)


dshiga:
Ran the mintegration test locally in dev and it succeeded:
https://job-manager.caas-dev.broadinstitute.org/jobs/3328273a-3262-4e89-a2d6-141559cb214d

dshiga:
Ran the mintegration test locally in dev and it's passing: 3328273a-3262-4e89-a2d6-141559cb214d

rexwangcc:
The Mintegration test failed because: http://ingest.integration.data.humancellatlas.org/submissions/5b33a58aba6b010007001f29 is still using the old URN instead of the new one (s3://org-humancellatlas-upload-dev/xxxxxxx/)

so we had the following error when staging files:
```
There is no upload area selected.
Please select one with "/usr/local/bin/hca upload select <uri_or_alias>"
```

I guess this PR need to wait until Ingest gets time to promote their dev to integration environment. @rolando-ebi

dshiga:
retest this please

pipeline_tools/http_requests.py
```diff
@@ -234,12 +234,16 @@ def _http_request(self, *args, **kwargs):
 
     @staticmethod
     def _get_method(http_method):
+        session = requests.Session()
+        # The default max_redirects is 30. We need to boost it because DSS sometimes redirects us more than 30 times,
+        # which causes an error if we don't allow for more.
+        session.max_redirects = 100
```
rexwangcc: These changes make sense to me. But I'm not sure why Blue is redirecting us is it based on any `User-Agent `headers? If so, how can we avoid that? Redirecting more than 30 times seems like a misbehavior to me.

rexwangcc:
Left one comment, otherwise looks good to me!

rolando-ebi:
@rexwangcc ingest integration is now up to date with the new CLI 👍

dshiga:
retest this please

dshiga:
retest this please

pipeline_tools/http_requests.py
dshiga: Blue box recently incorporated the checkout service into requests for the bundle manifest. This means that there is no longer a guarantee that they will instantly return the bundle manifest. Instead, they redirect us in a loop back to the same url (with some added parameters) so that when the manifest is ready we will get it. See @ttung's recently merged PR here: https://github.com/HumanCellAtlas/data-store/pull/1366. So we have to be able to handle a large number of redirects for these calls.

rexwangcc: Thanks for the explanation here! This makes more sense to me now.

PR #58, opened 2018-07-03
Make submit tasks idempotent

samanehsan:
- Check if analysis json is already in the submission envelope before adding it 
- Check if file has already been uploaded via the hca cli before uploading it 

- [x] Added or updated tests
- [x] Updated documentation

### DIFF
--- a/adapter_pipelines/submit.wdl
+++ b/adapter_pipelines/submit.wdl
@@ -142,10 +142,19 @@ task stage_files {
 
     # Stage the files
     files=( ${sep=' ' files} )
+
+    # Check if any files were already uploaded to the staging area
+    uploaded=($(hca upload list))
+
     for f in "$${lb}files[@]${rb}"
     do
-      echo "hca upload file $f"
-      hca upload file $f
+      file_path=($(echo "$f" | tr '/' '\n'))
+      file_name="$${lb}file_path[$${lb}#file_path[@]${rb} -1]${rb}"
+      is_uploaded=$(echo "$${lb}uploaded[@]${rb}" | grep -o $file_name | wc -w)
+      if [ $is_uploaded == 0 ]; then
+        echo "hca upload file $f"
+        hca upload file $f
+      fi
     done
   >>>
 
--- a/adapter_pipelines/submit.wdl
+++ b/adapter_pipelines/submit.wdl
@@ -140,21 +140,23 @@ task stage_files {
     echo "hca upload select $upload_urn"
     hca upload select $upload_urn
 
-    # Stage the files
-    files=( ${sep=' ' files} )
-
     # Check if any files were already uploaded to the staging area
-    uploaded=($(hca upload list))
+    uploaded_files=$(hca upload list)
 
-    for f in "$${lb}files[@]${rb}"
+    if [ -z $uploaded_files ]; then
+      unstaged_files=(${sep=' ' files})
+    else
+      echo "Checking which files to upload to the staging area..."
+      get-files-to-upload \
+        --files ${sep=' ' files} \
+        --uploaded_files $uploaded_files
+      unstaged_files=($(cat files.txt))
+    fi
+    # Stage the files
+    for f in "$${lb}unstaged_files[@]${rb}"
     do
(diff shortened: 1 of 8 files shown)


samanehsan:
retest this please

adapter_pipelines/submit.wdl
```diff
@@ -142,10 +142,19 @@ task stage_files {
...
+    # Check if any files were already uploaded to the staging area
+    uploaded=($(hca upload list))
+
     for f in "$${lb}files[@]${rb}"
     do
-      echo "hca upload file $f"
-      hca upload file $f
+      file_path=($(echo "$f" | tr '/' '\n'))
+      file_name="$${lb}file_path[$${lb}#file_path[@]${rb} -1]${rb}"
+      is_uploaded=$(echo "$${lb}uploaded[@]${rb}" | grep -o $file_name | wc -w)
```
dshiga: This is clever! But I think it would be clearer to write a bit of Python code that could filter the list down to just files that haven't been uploaded already, and then pass the filtered list to this loop instead. There would be an added advantage that you could write unit tests for that Python code, whereas the bash can't really be tested without running the WDL.

pipeline_tools/create_envelope.py
```diff
@@ -96,6 +101,35 @@ def create_submission_envelope(envelope_url, auth_headers, http_requests):
...
 
 
+def get_analysis_results(analyses_url, auth_headers, analysis_json_contents, http_requests):
+    """Checks the submission envelope for an analysis process with a protocol_id
+    that matches the analysis workflow id.
+
+    Args:
+        analyses_url (str): the url for creating the analysis record
+        auth_headers (dict): headers to use for auth
+        analysis_json_contents (dict): metadata describing the analysis
```
dshiga: Since you only use the workflow id, it would be clearer to pass only the workflow id in here and not the whole analysis json.

pipeline_tools/create_envelope.py
```diff
@@ -39,7 +40,10 @@ def run(submit_url, analysis_json_path, analysis_file_version):
     with open(analysis_json_path) as f:
         analysis_json_contents = json.load(f)
     analyses_url = get_subject_url(envelope_js, 'processes')
-    analysis_js = create_analysis(analyses_url, auth_headers, analysis_json_contents, http_requests)
+
+    analysis_js = get_analysis_results(analyses_url, auth_headers, analysis_json_contents, http_requests)
```
dshiga: Worth a comment explaining that we may have already created the analysis process in a previous attempt and that's why we're checking for it before creating it. Maybe rename the function get_analysis_process?

samanehsan:
retest this please

samanehsan:
retest this please

samanehsan:
retest this please

samanehsan:
retest this please

samanehsan:
retest this please

samanehsan:
retest this please

pipeline_tools/tests/data/analysis.json
```diff
@@ -1,96 +1,98 @@
...
-       "docker_image" : "https://uri-to-docker-image",
-       "cpus" : 8,
-       "memory" : "10GB",
-       "zone" : "us-east4-a"
-      }
+      "schema_type": "file"
+    }
+  ],
+  "protocol_core": {
+    "protocol_id": "ff84b09c-2b65-4d65-9ba2-7437e9ddac8b"
```
dshiga: I'm not sure what the current state of the schema is, but I thought we were going to have the protocol id be the same for all submissions from the same version of the pipeline, whereas the process id would be unique to the workflow run.

samanehsan: For the schema `https://schema.humancellatlas.org/type/protocol/analysis/7.0.0/analysis_protocol`, the analysis process was just renamed "analysis protocol" so the process_id became the protocol_id. There is a PR in the metadata repo (https://github.com/HumanCellAtlas/metadata-schema/pull/396)  that will fix the schema to include both an analysis protocol and analysis process schema.

PR #61, opened 2018-07-16
Follow retry_after headers from DSS 

samanehsan:
When making a request to the DSS for a bundle manifest, the request is redirected until the file is checked out into a bucket that we can access. The number of redirects can easily exceed the default maximum of 30. Instead, make the request follow the "retry after" header so that it does not redirect as many times. 

- [ ] Added or updated tests
- [ ] Updated documentation

### DIFF
--- a/pipeline_tools/http_requests.py
+++ b/pipeline_tools/http_requests.py
@@ -18,6 +18,12 @@ INDIVIDUAL_REQUEST_TIMEOUT = 'INDIVIDUAL_REQUEST_TIMEOUT'
 
 
 class RetryPolicy(retry_utils.Retry):
+    """Wrapper around the urllib3.retry module that overrides which status codes should follow the retry_after header.
+
+    Attributes:
+        retry_after_status_codes (frozenset): Which status codes follow the retry_after header
+
+    """
     def __init__(self, retry_after_status_codes={301}, **kwargs):
         super(RetryPolicy, self).__init__(**kwargs)
         self.RETRY_AFTER_STATUS_CODES = frozenset(retry_after_status_codes)
@@ -245,6 +251,8 @@ class HttpRequests(object):
     @staticmethod
     def _get_method(http_method):
         session = requests.Session()
+        # Follow retry-after headers for 301 status codes to avoid exceeding max_redirects (defaults to 30)
+        # when requesting files from the Data Store.
         retry_policy = RetryPolicy()
         adapter = requests.adapters.HTTPAdapter(max_retries=retry_policy)
         session.mount('https://', adapter)
--- a/pipeline_tools/http_requests.py
+++ b/pipeline_tools/http_requests.py
@@ -3,6 +3,7 @@ import requests
 import os
 import glob
 from tenacity import retry, stop_after_attempt, stop_after_delay, wait_exponential, retry_if_exception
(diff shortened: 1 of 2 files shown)


pipeline_tools/http_requests.py
```diff
@@ -238,9 +245,10 @@ def _http_request(self, *args, **kwargs):
     @staticmethod
     def _get_method(http_method):
         session = requests.Session()
-        # The default max_redirects is 30. We need to boost it because DSS sometimes redirects us more than 30 times,
-        # which causes an error if we don't allow for more.
-        session.max_redirects = 100
+        retry_policy = RetryPolicy(read=5, status=5, status_forcelist=frozenset({500, 502, 503, 504}))
```
dshiga: I'm not sure we want to retry 5xx errors in this way since we would normally be retrying those with tenacity. We probably want to keep retrying all errors in tenacity and only use urllib3 retry mechanism to follow redirects.

samanehsan: That makes sense to me!

samanehsan:
retest this please

samanehsan:
retest this please

pipeline_tools/http_requests.py
```diff
@@ -16,6 +17,12 @@
 INDIVIDUAL_REQUEST_TIMEOUT = 'INDIVIDUAL_REQUEST_TIMEOUT'
 
 
+class RetryPolicy(retry_utils.Retry):
+    def __init__(self, retry_after_status_codes={301}, **kwargs):
+        super(RetryPolicy, self).__init__(**kwargs)
+        self.RETRY_AFTER_STATUS_CODES = frozenset(retry_after_status_codes | retry_utils.Retry.RETRY_AFTER_STATUS_CODES)
```
dshiga: It looks like `RETRY_AFTER_STATUS_CODES` includes 503 and some 4xx errors. We probably just want `self.RETRY_AFTER_STATUS_CODES = frozenset(retry_after_status_codes)` so we restrict to just retrying 301s here.

samanehsan: Oh, I see! Thanks for catching that!

pipeline_tools/http_requests.py
```diff
@@ -238,9 +245,10 @@ def _http_request(self, *args, **kwargs):
     @staticmethod
     def _get_method(http_method):
         session = requests.Session()
-        # The default max_redirects is 30. We need to boost it because DSS sometimes redirects us more than 30 times,
-        # which causes an error if we don't allow for more.
-        session.max_redirects = 100
+        retry_policy = RetryPolicy()
```
dshiga: Would be helpful to have a comment about why we need this adapter / retry policy stuff.

pipeline_tools/http_requests.py
```diff
@@ -16,6 +17,12 @@
 INDIVIDUAL_REQUEST_TIMEOUT = 'INDIVIDUAL_REQUEST_TIMEOUT'
 
 
+class RetryPolicy(retry_utils.Retry):
```
dshiga: Some docs here would be good, even if short.

samanehsan:
retest this please

samanehsan:
retest this please

samanehsan:
retest this please

PR #70, opened 2018-08-16
Update to new HCA metadata schema and bundle structure

rexwangcc:
This PR consists of 2 major parts:
1. Update the pipeline-tools library to use the [`metadata-api`](https://github.com/HumanCellAtlas/metadata-api) library so we don't have to concern about the metadata schema changes directly anymore. 
2. Update the pipeline-tools library to adapt to the lastest bundle structure, including:
    - the way we prepare inputs for the pipeline, and 
    - how we generates analysis JSON for Ingest service.

**Note:** While removing a lot of existing code, we decide to keep using our `HttpRequests` class to talk to DSS service instead of using the `dss_client` Python API provided by the [`metadata-api`](https://github.com/HumanCellAtlas/metadata-api) library for now.

**Note:** This PR only focuses on SmartSeq2 Pipeline and breaks the Optimus pipeline intentionally since the latter one is not even running in secondary-analysis service yet. But there will be another PR fixes the Optimus pipeline shortly.

**Note:**  Since the [`metadata-api`](https://github.com/HumanCellAtlas/metadata-api) library only supports v5+ hca metadata schemas, from now on, we are also dropping the support for v4 metadata schema.

**Note:** This PR attempts to migrate as many test cases from unittest to pytest.

---

- [x] Added or updated tests
- [x] Updated documentation


pipeline_tools/input_utils.py
```diff
@@ -1,335 +1,168 @@
...
-
-        version_prefix = get_version_prefix(schema_version)
-        if version_prefix >= 6:
-            protocol_uuid = dcp_utils.get_file_uuid(manifest_files, 'protocol.json')
-            protocol_json = dcp_utils.get_file_by_uuid(protocol_uuid, dss_url, http_requests)
-            sequencing_protocol = [protocol for protocol in protocol_json['protocols'] if 'sequencing_protocol' in protocol['content']['describedBy']]
-            sequencing_protocol_id = sequencing_protocol[0]['hca_ingest']['document_id']
 
-    return inputs_metadata_json, sample_id_file_json, schema_version, sequencing_protocol_id
+    print("Getting bundle manifest for id {0}, version {1}".format(bundle_uuid, bundle_version))
```
rexwangcc: If there is no specific reason to use `print`, I prefer to place all prints with logging throughout the pipeline-tools, any objections?

pipeline_tools/tests/test_dcp_utils.py
```diff
@@ -1,34 +1,30 @@
...
-            cls.ss2_manifest_files_v4 = dcp_utils.get_manifest_file_dicts(cls.ss2_manifest_json_v4)
         cls.DSS_URL = "https://dss.mock.org/v0"
         cls.FILE_ID = "test_id"
         cls.BUNDLE_UUID = "test_uuid"
         cls.BUNDLE_VERSION = "test_version"
         cls.AUTH_TOKEN = {
             "access_token": "test_token",
-            "expires_in": 86400,
-            "token_type": "Bearer"
+            "expires_in"  : 86400,
```
samanehsan: nit: please remove the space before the `:` here and on lines 24 and 108

pipeline_tools/dcp_utils.py
```diff
@@ -28,8 +28,22 @@ def get_file_by_uuid(file_id, dss_url, http_requests):
 
 def get_manifest(bundle_uuid, bundle_version, dss_url, http_requests):
     """Retrieve manifest JSON file for a given bundle uuid and version.
+
     Retry with exponentially increasing wait times between requests if there are any failures.
 
+    TODO: Reduce the number of lines of code by switching to use DSS Python API client.
```
samanehsan: Can you please make a ticket for this?

samanehsan: Also can you fix the note about the return value? I know we'll replace this function soon, but I found this confusing! https://github.com/HumanCellAtlas/pipeline-tools/blob/9585c5c3a696bbdac94bdb070bbc2222b97bc0e0/pipeline_tools/dcp_utils.py#L53

pipeline_tools/input_utils.py
```diff
@@ -1,335 +1,168 @@
...
+    and this won't be required after migrating to use the HCA DSS Python API `dss_client` directly.
 
     Args:
-        lanes (dict): lane metadata
-        manifest_files (dict): file metadata
-
-    Returns:
-        Three lists of urls, representing fastqs for r1, r2, and i1, respectively.
-    In each list, the first item is for the first lane, the second item is for the second lane, etc.
+        item (typing.ItemsView): A dictionary's ItemsView object consisting of file_name and the manifest_entry.
```
samanehsan: Is `manifest_entry` the same as the `manifest` json file? If so, let's use the same name for consistency

pipeline_tools/input_utils.py
```diff
@@ -1,335 +1,168 @@
...
-    r1 = [manifest_files['name_to_meta'][lane['r1']]['url'] for lane in lanes]
-    r2 = [manifest_files['name_to_meta'][lane['r2']]['url'] for lane in lanes]
-    i1 = [manifest_files['name_to_meta'][lane['i1']]['url'] for lane in lanes]
+    file_name, manifest_entry = item
+    file_uuid = manifest_entry['uuid']
+    return file_name, dcp_utils.get_file_by_uuid(file_id=file_uuid, dss_url=dss_url, http_requests=http_requests)
 
-    return r1, r2, i1
 
+def get_metadata_files(metadata_files_dict, dss_url, num_workers=None):
```
samanehsan: This can be for future work, but there should be a way for us to get the bundle with the file metadata included in it already. Then once the bundles are plate based and include many more files, we won't have to fetch each one individually!

pipeline_tools/input_utils.py
samanehsan: Oh nevermind! I see that now that the `manifest_entry` is the info about a particular item in the `manifest` json file

samanehsan:
This looks great @rexwangcc! Can you add some tests for the new helper functions? Also since this PR removes python 2.7 support, please update the `.travis.yml`: https://github.com/HumanCellAtlas/pipeline-tools/blob/master/.travis.yml#L3

pipeline_tools/dcp_utils.py
rexwangcc: Good catch, I will fix the incorrect docstring here. Also, will make a ticket for this!

pipeline_tools/input_utils.py
rexwangcc: Good point! Actually it is already there: https://github.com/HumanCellAtlas/metadata-api/blob/develop/examples/simple.ipynb 

As I said in the PR's description, we cannot use that approach(dss_client.download_bundle_metadata) because we have the `HttpRequests` class which helps us log the requests and retry on various failures. After discussing with @dshiga I think it is better we leave it as a future work and refactor the pipeline-tools to make it more manageable and efficient.

pipeline_tools/input_utils.py
rexwangcc: yep, before going to JSON, it's now a class defined in "metadata-api"

rexwangcc:
@samanehsan Thanks for taking a look at this WIP PR! the update for `.travis.yml` sounds great, will do!

requirements.txt
```diff
@@ -1,4 +1,5 @@
 requests==2.18.4
 google-cloud-storage==1.8.0
 tenacity==4.10.0
-git+git://github.com/broadinstitute/cromwell-tools.git@v0.3.1
+git+git://github.com/broadinstitute/cromwell-tools.git@v0.5.0
+git+git://github.com/HumanCellAtlas/metadata-api@master#egg=hca-metadata-api[dss]
```
samanehsan: We should pin to a specific commit or tag here in case breaking changes are merged into the master branch.

pipeline_tools/get_analysis_metadata.py
```diff
@@ -1,11 +1,21 @@
...
 
 
-def get_workflow_id(analysis_output_path):
+def _get_caas_backend():
+    """Get the url to the Cromwell-as-a-service backend.
+
+    This is a temporary function abstracted from the other functions which are hard-coding the url. FIXME: it doesn't
+    make sense to hard-code the CaaS url anymore once we migrate to use caas-prod as backend.
+    """
+    return 'https://cromwell.caas-dev.broadinstitute.org/api/workflows/v1'
```
samanehsan: Since this returns a constant value, how about setting a variable instead, e.g: 

`CAAS_URL = 'https://cromwell.caas-dev.broadinstitute.org/api/workflows/v1'`

I don't like the fact that this is hard-coded either, since we will have to change this to the caas-prod URL in addition to changing the Lira config. But the alternative involves a lot of code changes to parameterize this, and I don't expect this to change often so I think it's okay to keep it defined here.

pipeline_tools/create_analysis_metadata.py
```diff
@@ -0,0 +1,502 @@
...
+                'stop_time'   : task['end'],
+                'log_out'     : task['stdout'],
+                'log_err'     : task['stderr']
+            }
+            output_tasks.append(out_task)
+    sorted_output_tasks = sorted(output_tasks, key=lambda k: k['task_name'])
+    return sorted_output_tasks
+
+
+def get_format(path, extension_to_format):
```
samanehsan: I suggest renaming this to `get_file_format` to make it clearer what this function is doing

pipeline_tools/get_analysis_metadata.py
```diff
@@ -18,19 +28,81 @@ def get_workflow_id(analysis_output_path):
...
+        caas_key_file (str): path to CaaS service account JSON key file.
+
+    Raises:
+        requests.HTTPError: for 4xx errors or 5xx errors beyond the timeout
+    """
+
+    def log_before(workflowId):
+        print('Getting the version for adapter workflow {}'.format(workflowId))
+
+    if use_caas:
```
samanehsan: You can move some of this repeated code that gets the auth, headers and cromwell_url to a separate function or use the ones in cromwell_tools

samanehsan:
Please update the inputs to the stub submission wdl as well, otherwise the mintegration test will fail for skylab: https://github.com/HumanCellAtlas/pipeline-tools/blob/master/adapter_pipelines/submit_stub/submit.wdl

pipeline_tools/get_analysis_metadata.py
dshiga: I think we should pass the cromwell url in as a parameter where it's needed, and not hard code it in pipeline-tools, even temporarily.

pipeline_tools/get_analysis_metadata.py
```diff
@@ -18,19 +28,81 @@ def get_workflow_id(analysis_output_path):
...
+    url = analysis_output_path
+    calls = url.split('/call-')
+    adapter_workflow_id = calls[0].split('/')[-1]
+    print('Got adapter workflow UUID: {0}'.format(adapter_workflow_id))
+    with open('adapter_workflow_id.txt', 'w') as f:
+        f.write(adapter_workflow_id)
+    return adapter_workflow_id
+
+
+def get_adapter_workflow_version(runtime_environment,
```
dshiga: I don't think we should be making a call to Cromwell for this. It adds another external dependency and possible point of failure that I'd rather not have. It would simplify things if we hard code a version parameter in each adapter WDL or analysis WDL and pass it to the task that needs it, e.g. `String pipeline_version = v1.2.3`.

pipeline_tools/create_envelope.py
```diff
@@ -1,277 +1,361 @@
...
-    """Create submission in ingest service.
+def get_subject_url(target_dic, subject):
+    """Get the Ingest service url for a given subject.
 
     Args:
-        submit_url (str): url of ingest service to use
-        analysis_json_path (str): path to analysis json file
-        schema_url (str): URL for retrieving HCA metadata schemas
-        analysis_file_version (str): version of metadata schema that output files should conform to
+        target_dic (dict): Dict representing the JSON response from the root Ingest service url.
```
dshiga: `endpoint_dict` would be better

pipeline_tools/create_envelope.py
```diff
@@ -1,277 +1,361 @@
...
-            }
-        }
-        output_refs.append(output_ref)
-    return output_refs
+    print('Adding file: {0} to the file reference.'.format(file_ref['fileName']))
+    response = http_requests.put(file_refs_url, headers=auth_headers, json=file_ref)
+    return response.json()
+
+
+def build_envelope(submit_url, analysis_protocol_path, analysis_process_path, raw_schema_url,
```
dshiga: I find it easier to read when the top-level function (`build_envelope` in this case), is at the top of the file, so that the first thing I read is more of an overview, and then I can scroll down to see the details / helper functions.

pipeline_tools/input_utils.py
```diff
@@ -1,219 +1,181 @@
...
-
+    file_name, manifest_entry = item
+    file_uuid = manifest_entry['uuid']
+    return file_name, dcp_utils.get_file_by_uuid(file_id=file_uuid, dss_url=dss_url, http_requests=http_requests)
 
-def _get_sample_id_file_uuid_v4(manifest_files):
-    """Get the uuid of the assay.json file"""
-    return dcp_utils.get_file_uuid(manifest_files, 'assay.json')
 
+def get_metadata_files(metadata_files_dict, dss_url, num_workers=None):
```
dshiga: Rather than downloading the metadata as individual files, could we just make one call to query blue box to return all metadata as one big json blob?

pipeline_tools/create_envelope.py
```diff
@@ -1,277 +1,361 @@
...
-    return envelope_js
+    response = http_requests.post(envelope_url, '{}', headers=auth_headers)
+    envelope_dict = response.json()
+    return envelope_dict
 
 
-def get_analysis_process(analyses_url, auth_headers, analysis_workflow_id, http_requests):
-    """Checks the submission envelope for an analysis process with a protocol_id
-    that matches the analysis workflow id.
+# FIXME: questionable in new bundle structure :(
```
dshiga: What does this mean?

rexwangcc: This is a to-do placeholder for me 😄 Since we are separating analysis_json into analysis_protocol and analysis_process now, we have to make an additional call like `curl -X PUT -H “Content-Type: text/uri-list” -d “http://api.ingest.integration.data.humancellatlas.org/protocols/xxx” “http://api.ingest.integration.data.humancellatlas.org/processes/xxx/protocols”`

And I'm still working on that.

pipeline_tools/input_utils.py
rexwangcc: Based on an offline discussion with @dshiga The current function is copied over from the [`metadata-api`](https://github.com/HumanCellAtlas/metadata-api/blob/1b7192cecbef43b5befecc4153bf2e2f4db5bb16/src/humancellatlas/data/metadata/helpers/dss.py#L14), I will make a ticket for this as a future work. The comments here totally make sense, otherwise we can potentially run into some capabilities problems in the future. Also, `metadata-api` repo might be a better palce to handle this logic, once we start to use the `dss_client`

pipeline_tools/create_envelope.py
rexwangcc: Yes, that totally makes sense, I wrote this function at last during development, will push it to the top!

pipeline_tools/create_envelope.py
rexwangcc: changed, thanks!

pipeline_tools/get_analysis_metadata.py
rexwangcc: Per an offline discussion with @dshiga and also @jishuxu  I will make a PR in Skylab for this. changing both ss2 and Optimus. I agree that we cannot really rely on the workflow labels for everything.

pipeline_tools/create_analysis_metadata.py
```diff
@@ -0,0 +1,502 @@
...
+    Returns:
+        hex_str (str): A string giving the hexadecimal-encoded representation of the given data.
+    """
+    base64_bytes = base64.b64decode(base64_str.strip('\n"\''))
+    hex_bytes = codecs.encode(base64_bytes, 'hex')
+    hex_str = hex_bytes.decode('utf-8')
+    return hex_str
+
+
+def get_analysis_describedBy(schema_url, schema_type, schema_version):
```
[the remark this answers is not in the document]
rexwangcc: changed, thanks!

pipeline_tools/create_analysis_metadata.py
```diff
@@ -0,0 +1,502 @@
...
+    Returns:
+        analysis_protocol (dict): A dict representing the analysis_protocol json file to be submitted.
+    """
+    SCHEMA_TYPE = 'protocol'
+
+    analysis_protocol = {
+        'describedBy'         : get_analysis_describedBy(schema_url=raw_schema_url,
+                                                         schema_type=SCHEMA_TYPE,
+                                                         schema_version=analysis_protocol_schema_version),
+        'schema_type'         : SCHEMA_TYPE,
```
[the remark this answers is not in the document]
rexwangcc: To all this type of comments: the extra spaces here are intentional 🌞 Although they are violating the PEP 8, I found a lot of repos are adopting this style, and it can be turned on in PyCharm -> Tools -> Editor -> Code Style -> Python -> Other -> `Align on colon` Personally I feel it makes code more human readable, consider:

```python
'schema_type'        : SCHEMA_TYPE,
'process_core'       : get_analysis_process_core(analysis_workflow_id=analysis_id),
'process_type'       : get_analysis_process_type(),
'timestamp_start_utc': workflow_metadata.get('start'),
'timestamp_stop_utc' : workflow_metadata.get('end'),
'input_bundles'      : input_bundles_string.split(','),
'reference_bundle'   : reference_bundle,
'tasks'              : workflow_tasks,
'inputs'             : inputs,
```
compare to:

```python
'schema_type' : SCHEMA_TYPE,
'process_core': get_analysis_process_core(analysis_workflow_id=analysis_id),
'process_type': get_analysis_process_type(),
'timestamp_start_utc': workflow_metadata.get('start'),
'timestamp_stop_utc' : workflow_metadata.get('end'),
'input_bundles': input_bundles_string.split(','),
'reference_bundle': reference_bundle,
'tasks': workflow_tasks,
'inputs': inputs,
```

But, this is ONLY my personal preferences, I'm happy to change it if we'd reach some agreements on code style (or just using a linter like we did in jmui)

pipeline_tools/create_analysis_metadata.py
rexwangcc: Will do!

pipeline_tools/get_analysis_metadata.py
rexwangcc: This was copied over, but I will make it as a parameter in static inputs as suggested.

pipeline_tools/get_analysis_metadata.py
rexwangcc: Will look into it, but if it ends up being a lot more work, would you mind me leaving this as another ticket?

pipeline_tools/get_analysis_metadata.py
```diff
@@ -18,19 +19,81 @@ def get_workflow_id(analysis_output_path):
...
+        requests.HTTPError: for 4xx errors or 5xx errors beyond the timeout
+    """
+
+    def log_before(workflowId):
+        print('Getting the version for adapter workflow {}'.format(workflowId))
+
+    cromwell_url = cromwell_url
+
+    if use_caas:
+        json_credentials = caas_key_file or "/cromwell-metadata/caas_key.json"
```
samanehsan: @rexwangcc can you update this to use the right caas-key depending on the cromwell url? E.g. caas-prod should use the caas-prod key. Right now, there's only one "caas_key" in the docker image that is for caas dev so running pipelines through caas-prod will fail 😞

adapter_pipelines/submit.wdl
```diff
@@ -23,15 +24,17 @@ task get_metadata {
...
       --use_caas ${use_caas}
   >>>
   runtime {
-    docker: "gcr.io/broad-dsde-mint-${runtime_environment}/cromwell-metadata:v0.19.0"
+    docker: "gcr.io/broad-dsde-mint-${runtime_environment}/cromwell-metadata:cw-update-to-new-bundle-structure"
   }
   output {
     File metadata = "metadata.json"
     String workflow_id = read_string("workflow_id.txt")
+    String adapter_workflow_id = read_string("adapter_workflow_id.txt")
```
samanehsan: Can we remove this output file? It doesn't look like it's used anywhere else in the submit wdl.

pipeline_tools/get_analysis_metadata.py
```diff
@@ -18,19 +19,81 @@ def get_workflow_id(analysis_output_path):
...
+        http_requests: `http_requests.HttpRequests` instance, a wrapper around requests provides better retry and
+                       logging.
+        use_caas (bool): whether or not to use Cromwell-as-a-Service.
+        caas_key_file (str): path to CaaS service account JSON key file.
+
+    Raises:
+        requests.HTTPError: for 4xx errors or 5xx errors beyond the timeout
+    """
+
+    def log_before(workflowId):
```
samanehsan: please change this to `workflow_id` for consistency

pipeline_tools/tests/data/analysis_protocol.json
```diff
@@ -0,0 +1,11 @@
+{
+  "computational_method": "SmartSeq2SingleCell",
+  "describedBy": "http://schema.integration.data.humancellatlas.org/type/protocol/analysis/7.1.0/analysis_protocol",
```
samanehsan: We're using analysis_protocol version 8.0.0 now right? Does that require updating this test data as well?

pipeline_tools/create_analysis_metadata.py
```diff
@@ -0,0 +1,502 @@
...
+                            output_url_to_md5,
+                            extension_to_format,
+                            analysis_file_version,
+                            run_type):
+    """Collect and create the information about the analysis process for submission to Ingest service.
+
+    Based the design of this HCA metadata type, analysis_process will vary between each analysis run, even if they share
+    the same version of pipeline. Most of the content of this process will be collected from the Cromwell workflow
+    metadata. This function strictly follows the metadata schema defined at:
+    https://github.com/HumanCellAtlas/metadata-schema
```
samanehsan: This should point to the json schema the function is using, although it's only on the integration branch at the moment: https://github.com/HumanCellAtlas/metadata-schema/blob/integration/json_schema/type/process/analysis/analysis_process.json

pipeline_tools/create_analysis_metadata.py
```diff
@@ -0,0 +1,502 @@
...
+    SCHEMA_TYPE = 'process'
+
+    workflow_metadata = get_workflow_metadata(metadata_file)
+    workflow_tasks = get_workflow_tasks(workflow_metadata)
+
+    analysis_process = {
+        'describedBy'        : get_analysis_described_by(schema_url=raw_schema_url,
+                                                         schema_type=SCHEMA_TYPE,
+                                                         schema_version=analysis_process_schema_version),
+        'schema_type'        : SCHEMA_TYPE,
```
samanehsan: I prefer the style without the extra spaces 🙂

pipeline_tools/get_analysis_metadata.py
rexwangcc: Once we start to use caas-prod, we won't use `caas-dev` anymore, so I think we should only have 1 caas_key.json instead of differentiating caas-prod and caas-dev in pipeline-tools 's code base, which will require us to parse the url and check prod/dev in the parsed strings. Do you mind I leave this (updating the service account in cromwell-metadata docker image) as a future work, along with other works that migrating to caas-prod? If so, the only thing we need to do there is rebuild the image and push it to gcr, no need to change anything in pipeline-tools

samanehsan: Sure, that makes sense to me

pipeline_tools/create_envelope.py
```diff
@@ -1,277 +1,416 @@
...
-            'schema_type': 'file',
-            'file_core': {
-                'file_name': file_name,
-                'file_format': out['file_core']['file_format']
-            }
-        }
-        output_refs.append(output_ref)
-    return output_refs
+    link_headers = {'content-type': 'text/uri-list'}
+    response = http_requests.put(link_url, headers=link_headers, data=analysis_protocol_url)
```
samanehsan: Can you return `response.json()` at the end of this function?

pipeline_tools/create_envelope.py
```diff
@@ -1,277 +1,416 @@
 #!/usr/bin/env python
-
-import json
 import argparse
-import requests
-from .dcp_utils import get_auth_token, make_auth_header
+import json
+from typing import List
```
samanehsan: This import is not being used, can we remove it?

rexwangcc: In fact, it is used in some docstrings, like this: https://github.com/HumanCellAtlas/pipeline-tools/pull/70/files/39bc791d38668bd5f0ba07e7a22edb9509793ac5#diff-c5363cf30132023e251443047cab1a17R334 , PyCharm/IntelliJ/IDEs are using it to give warnings, and pytest is also using it to make some typing assertions.

samanehsan: Thanks, I see where it's used now. So it's required as an import even if it's used in a docstring/comment?

pipeline_tools/create_envelope.py
rexwangcc: No I cannot :=( The response from Ingest for this endpoint currently returns non-JSON messages, `response.json()` will throw an error about it. Since the `response` is not used by anywhere else, I can replace it with `http_requests.put(link_url, headers=link_headers, data=analysis_protocol_url)` so it's less confusing?

pipeline_tools/create_envelope.py
rexwangcc: Yup, it's required. If we want to adopt type annotations in the future, it will be even more explicitly required.

pipeline_tools/tests/data/analysis_protocol.json
rexwangcc: Good catch, I will update the testing data shortly, thanks!

adapter_pipelines/submit.wdl
rexwangcc: Sure, will remove it.

pipeline_tools/get_analysis_metadata.py
rexwangcc: will do

pipeline_tools/create_analysis_metadata.py
rexwangcc: will do!

pipeline_tools/create_envelope.py
```diff
@@ -1,277 +1,416 @@
...
-                                  '{}',
-                                  headers=auth_headers)
-    envelope_js = response.json()
-    return envelope_js
+    response = http_requests.post(envelope_url, '{}', headers=auth_headers)
+    envelope_dict = response.json()
+    return envelope_dict
+
+
+# FIXME: questionable in new bundle structure :(
```
samanehsan: @rexwangcc we can confirm that this is correct by checking what the new submission envelope looks like

rexwangcc: Sorry for the confusion, I have removed the fixmes in a newer commit, but haven't push them yet. Nice catch!

pipeline_tools/create_envelope.py
samanehsan: Ok, thanks for the explanation!

rexwangcc: Thank you!

pipeline_tools/create_envelope.py
samanehsan: Ahh okay I wonder if that was on purpose? 🙁

rexwangcc: I have no idea 🤔 maybe @aaclan-ebi or @rolando-ebi can confirm?

aaclan-ebi: yup, that API is provided by Spring Data REST, and and the response for its result will be a HTTP status 204 (no content).

rexwangcc: @aaclan-ebi Cool, thanks for the explanation here!

rexwangcc:
This PR passed the mintegration test locally, workflow is `0f9e7bc3-0134-46c9-b93f-8983d7403c92` in Caas-dev. In order to make it pass the Mintegration test here, we have to fix the mintegration test by using a new bundle later. As discussed with @samanehsan offline, I'm going to merge this PR after changing the docker tags in adapter workflows.

rexwangcc:
Merging now.

PR #76, opened 2018-09-06
By default, don't submit md5 checksums in metadata (287)

dshiga:
Submitting md5s at scale is not working well, causing a lot of failures in the scale test, and long term there are other reasons why we might not want to continue doing it. For now, this PR turns md5s off by default but adds a parameter to turn them back on if we want to.

See HumanCellAtlas/secondary-analysis#287

- [x] Added or updated tests
- [x] Updated documentation

### DIFF
--- a/adapter_pipelines/submit.wdl
+++ b/adapter_pipelines/submit.wdl
@@ -62,6 +62,7 @@ task create_submission {
   Int? individual_request_timeout
   Boolean record_http
   String pipeline_tools_version
+  Boolean add_md5s
 
   command <<<
     export RECORD_HTTP_REQUESTS="${record_http}"
@@ -92,7 +93,8 @@ task create_submission {
       --analysis_file_version ${analysis_file_version} \
       --inputs_file ${write_objects(inputs)} \
       --outputs_file ${write_lines(outputs)} \
-      --format_map ${format_map}
+      --format_map ${format_map} \
+      --add_md5s ${add_md5s}
 
     # Now build the submission object
     create-envelope \
@@ -244,6 +246,7 @@ workflow submit {
   # By default, don't record http requests
   Boolean record_http = false
   String pipeline_tools_version
+  Boolean add_md5s
 
   call get_metadata {
     input:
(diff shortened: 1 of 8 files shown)


rexwangcc:
Also, please update the unit tests 😄

pipeline_tools/create_analysis_metadata.py
```diff
@@ -132,6 +129,21 @@ def get_inputs(inputs_file):
     return inputs
 
 
+def get_outputs(output_urls, extension_to_format, schema_url, analysis_file_version):
```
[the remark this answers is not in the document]
dshiga: Yeah, I'm in the process of refactoring and pushing commits as I go, so it might be a bit incoherent until the refactoring is complete.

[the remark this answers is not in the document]
rexwangcc: I see, I should wait for now! ⏰

pipeline_tools/create_analysis_metadata.py
```diff
@@ -448,25 +459,38 @@ def main():
...
     with open(args.outputs_file) as f:
         output_urls = f.read().splitlines()
-    client = storage.Client()
-    output_url_to_md5 = get_md5s(output_urls, client)
+    outputs = get_outputs(output_urls=output_urls,
+                          extension_to_format=extension_to_format,
+                          schema_url=schema_url,
+                          analysis_file_version=args.analysis_file_version)
+    print('The content of outputs are: ')
+    print(outputs)
```
rexwangcc: nit: Move the above after `if args.add_md5s == 'true'` block so that if we enable the md5s, it will print the right outputs to the console.

pipeline_tools/create_analysis_metadata.py
```diff
@@ -448,25 +459,38 @@ def main():
...
 
-    # Get inputs and add MD5 checksum to them
-    inputs = get_inputs(args.inputs_file)
-    input_urls = get_input_urls(inputs)
-    input_url_to_md5 = get_md5s(input_urls, client)
-    inputs = add_md5s(inputs, input_url_to_md5)
+    # Add md5 checksums to input and output metadata if needed
```
rexwangcc: Could you briefly explain the scaling problem with `md5s` here in comments, just as a record?

adapter_pipelines/submit.wdl
```diff
@@ -62,6 +62,7 @@ task create_submission {
   Int? individual_request_timeout
   Boolean record_http
   String pipeline_tools_version
+  Boolean add_md5s
```
rexwangcc: Should we also add this parameter to the stub submit wdl? Otherwise it may fail PRs in skylab i guess

PR #78, opened 2018-09-11
Retry HTTP requests on 409 errors

samanehsan:
Retry 409 errors to handle the optimistic lock exception thrown by the Ingest API when linking the analysis process to protocol when running many submissions simultaneously: https://github.com/HumanCellAtlas/ingest-central/issues/38

- [ ] Added or updated tests
- [ ] Updated documentation

### DIFF
--- a/pipeline_tools/http_requests.py
+++ b/pipeline_tools/http_requests.py
@@ -159,7 +159,12 @@ class HttpRequests(object):
             print('{0} {1}'.format(now, repr(error)))
 
             def is_retryable_status_code(error):
-                return isinstance(error, requests.HTTPError) and not (400 <= error.response.status_code <= 499 and error.response.status_code != 409)
+                if not isinstance(error, requests.HTTPError):
+                    return False
+                if error.response.status_code == 409:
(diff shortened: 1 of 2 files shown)


rexwangcc:
retest this please

samanehsan:
retest this please

pipeline_tools/http_requests.py
```diff
@@ -159,7 +159,7 @@ def is_retryable(error):
             print('{0} {1}'.format(now, repr(error)))
 
             def is_retryable_status_code(error):
-                return isinstance(error, requests.HTTPError) and not (400 <= error.response.status_code <= 499)
+                return isinstance(error, requests.HTTPError) and not (400 <= error.response.status_code <= 499 and error.response.status_code != 409)
```
dshiga: Hmm, this condition is getting a bit long and confusing.

What about something like:
```
if not isinstance(error, requests.HTTPError):
    return False
if error.response.status_code == 409:
    return True
else:
    return not (400 <= error.response.status_code <= 499)
```

PR #86, opened 2018-10-09
Add web_summary.html to CellRanger adapter outputs

samanehsan:
- [ ] Added or updated tests
- [ ] Updated documentation


samanehsan:
Tested locally and confirmed that this change still works with both the gencode and ensembl references for the cellranger pipeline

PR #90, opened 2018-10-12
Rename cellranger outputs so that they are unique

samanehsan:
Since the HCA upload service overwrites files that have the same name (even if the file paths are unique), change the file names output from cellranger so that they are unique.

### DIFF
--- a/adapter_pipelines/cellranger/adapter.wdl
+++ b/adapter_pipelines/cellranger/adapter.wdl
@@ -243,7 +243,7 @@ workflow Adapter10xCount {
 
   # Rename analysis files so that all the file names are unique. For example, rename
   # "${sample_id}/outs/raw_gene_bc_matrices/${reference}/barcodes.tsv" to "raw_barcodes.tsv" so that
-  # it does not overwrite "${sample_id}/outs/raw_gene_bc_matrices/${reference}/barcodes.tsv"
+  # it does not overwrite "${sample_id}/outs/filtered_gene_bc_matrices/${reference}/barcodes.tsv"
   # when uploading files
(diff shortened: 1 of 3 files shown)


adapter_pipelines/cellranger/adapter.wdl
```diff
@@ -216,12 +241,23 @@ workflow Adapter10xCount {
       pipeline_tools_version = pipeline_tools_version
   }
 
+  # Rename analysis files so that all the file names are unique. For example, rename
+  # "${sample_id}/outs/raw_gene_bc_matrices/${reference}/barcodes.tsv" to "raw_barcodes.tsv" so that
+  # it does not overwrite "${sample_id}/outs/raw_gene_bc_matrices/${reference}/barcodes.tsv"
```
rexwangcc: I'm not sure I understand the comments here, so we have multiple `"${sample_id}/outs/raw_gene_bc_matrices/${reference}/barcodes.tsv"` files? If so, how could renaming them to `"raw_barcodes.tsv"` solves the problem?

samanehsan: Oh I made a mistake copying the path! 

There is one file named "${sample_id}/outs/**filtered_gene_bc_matrices**/${reference}/barcodes.tsv"and another named "${sample_id}/outs/**raw_gene_bc_matrices**/${reference}/barcodes.tsv" and since they are both called "barcodes.tsv", this task renames the second one to "raw_barcodes.tsv"

rexwangcc: Ah, that makes more sense!

PR #94, opened 2018-10-22
Handle 10x data with no I1 file

samanehsan:
**Purpose:**
This fixes HumanCellAtlas/secondary-analysis#392, where the `GetInputs` task of the CellRanger Adapter WDL output `Array[File] i1_fastq = [""]` for bundles with no I1 files. This is an invalid file path, which causes issues when downstream tasks try to localize it.

**Changes:**
The `GetInputs` task outputs a file that lists all fastq urls and another file containing the CellRanger-compatible fastq file names. Use the general `RenameFiles` task to rename the fastq files instead of doing a scatter over RenameFastqFile, which will fail if the I1 file is an empty string.

**Testing:**
This bundle in the integration data store has no `I1.fastq.gz` file: `c356007c-53a1-41a1-b6b8-b378d86b1f78.2018-10-22T171055.307092Z`

Running the mintegration test locally using this bundle passes: https://job-manager.caas-prod.broadinstitute.org/jobs/b9db7a7f-5582-45a9-b321-47e14432adc6

The mintegration test in jenkins uses a 10x bundle with an I1 file, which also passes ✅


pipeline_tools/input_utils.py
```diff
@@ -218,3 +219,59 @@ def create_optimus_input_tsv(uuid, version, dss_url):
         f.write('{0}'.format(sample_id))
 
     print('Finished writing files')
+
+
+def get_cellranger_inputs(uuid, version, dss_url):
```
rexwangcc: nit: to align with other functions, could we name this to something like: `create_cellranger_input_files`?

pipeline_tools/input_utils.py
```diff
@@ -218,3 +219,59 @@ def create_optimus_input_tsv(uuid, version, dss_url):
...
+    Args:
+        uuid (str): the bundle uuid
+        version (str): the bundle version
+        dss_url (str): the DCP Data Storage Service
+
+    Returns:
+        None
+
+    Raises:
+        optimus_utils.LaneMissingFileError if any fastqs are missing
```
rexwangcc: nit: `optimus_utils.LaneMissingFileError: if any fastqs are missing` (the colon)

adapter_pipelines/cellranger/adapter.wdl
```diff
@@ -190,39 +165,29 @@ workflow Adapter10xCount {
...
-      input:
-        r1 = GetInputs.r1_fastq[i],
-        r2 = GetInputs.r2_fastq[i],
-        i1 = GetInputs.i1_fastq[i],
-        sample_id = GetInputs.sample_id,
-        lane = GetInputs.lanes[i],
-        pipeline_tools_version = pipeline_tools_version
-      }
-    }
+  # Rename the fastq files to the format required by CellRanger:
```
rexwangcc: If I understand it correctly, cellranger doesn't support scatter tasks anyway, so we could remove the `lane` parameter here right?

pipeline_tools/input_utils.py
rexwangcc: By the way, are there any reasons stop us from using "TSV"? I guess we could consolidate the 2 txt files into one if we use tsv?

adapter_pipelines/cellranger/adapter.wdl
samanehsan: We still need to rename the fastq files, but now this is passing in an array of files instead of doing a scatter

pipeline_tools/input_utils.py
samanehsan: I created the `txt` files to be consistent with the behavior for Optimus, which...even though it does not create a tsv file is named `create_optimus_input_tsv`

We could use a tsv file here instead, but I think this is more readable. Also, we may want to change this approach in the future so that we are downloading the fastq files in `GetInputs` instead of relying on the fragile String --> File conversion with `read_lines` so I prefer to leave this as it is for now.

PR #99, opened 2018-10-26
Add monitor script for cellranger and update cromwell-metadata image

rexwangcc:
- [ ] Added or updated tests
- [ ] Updated documentation


samanehsan:
Since the cromwell-metadata docker base image changed, how about updating the docker tag with the new image version?

PR #106, opened 2018-11-30
Remove unused gtf_file parameter from SS2 adapter

mckinsel:

PR #109, opened 2018-12-10
Remove unused gtf_file parameter from SS2 adapter

mckinsel:
This is the same as #106, but with the branch name changed to match the one in HumanCellAtlas/skylab#149 so jenkins will test them together.


samanehsan:
I think the integration test is broken again 😭

mckinsel:
This seems to have an permanently broken jenkins test, but an identical PR, #145, is passing. So just close this one and go with that one 🤷‍♂

PR #117, opened 2019-02-11
Standardize Cromwell timestamps to follow analysis JSON schema

samanehsan:
### Purpose
Fixes https://github.com/HumanCellAtlas/secondary-analysis/issues/523

Zenhub ticket: https://app.zenhub.com/workspaces/dcp-backlogs-5ac7bcf9465cb172b77760d9/issues/humancellatlas/secondary-analysis/523

---
### Changes
Format timestamps from Cromwell metadata to follow the date-time JSON schema format when constructing the analysis JSON information.

When converting the timestamps into a datetime object, the `arrow` python library does not require a format string (the python standard datetime library does) so it is used here to better handle a variety of formats.

---
### Review Instructions

---
### PR Checklist
_Please ensure the following when opening a PR:_

- [x] This PR added or updated tests.
- [x] This PR updated docstrings or documentation.
- [x] This PR applied code style guidelines:
    - [Google style docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google) Python code.
    - Mint WDL style guide for WDL.
- [x] This PR considered generalizability beyond our own use case.

---
### Follow-up Discussions

Currently, the `start` and `end` times for the analysis workflow will default to `None` if they are not present in the metadata: https://github.com/HumanCellAtlas/pipeline-tools/blob/master/pipeline_tools/create_analysis_metadata.py#L63

However, the resulting `null` value in the analysis process JSON schema will not validate in that case. Should it just fail instead?

### DIFF
--- a/adapter_pipelines/Optimus/adapter.wdl
+++ b/adapter_pipelines/Optimus/adapter.wdl
@@ -126,7 +126,7 @@ workflow AdapterOptimus {
   Int max_cromwell_retries = 0
   Boolean add_md5s = false
 
-  String pipeline_tools_version = "se-fix-timestamps"
+  String pipeline_tools_version = "v0.46.1"
 
   call GetInputs as prep {
     input:
--- a/adapter_pipelines/Optimus/adapter.wdl
+++ b/adapter_pipelines/Optimus/adapter.wdl
@@ -126,7 +126,7 @@ workflow AdapterOptimus {
   Int max_cromwell_retries = 0
   Boolean add_md5s = false
 
-  String pipeline_tools_version = "v0.46.0"
+  String pipeline_tools_version = "se-fix-timestamps"
 
   call GetInputs as prep {
     input:
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ setup(name='pipeline-tools',
       install_requires=[
           'arrow>=0.12.1',
           'google-auth>=1.6.1,<2',
-          'google-cloud-storage>=1.8.0,<2',
+          'google-cloud-storage>=1.10.0,<2',
           'hca>=4.5.0,<5',
           'hca-metadata-api',
           'mock>=2.0.0,<3',
@@ -25,7 +25,6 @@ setup(name='pipeline-tools',
           'requests-mock>=1.5.2,<2',
           'setuptools_scm>=2.0.0,<3',
           'tenacity>=4.10.0,<5',
-          'google-cloud-storage>=1.10.0,<2',
           'PyJWT==1.6.4'
       ],
       entry_points={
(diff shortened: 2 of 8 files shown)


samanehsan:
retest this please

setup.py
```diff
@@ -15,6 +15,7 @@
       license='BSD 3-clause "New" or "Revised" License',
       packages=['pipeline_tools'],
       install_requires=[
+          'arrow>=0.12.1',
           'google-auth>=1.6.1,<2',
           'google-cloud-storage>=1.8.0,<2',
```
pshapiro4broad: Is it intentional that this doesn't match the version text in `requirements.txt`? I'm not familiar with python but I'd hope there would be a way to only specify the dependency list once, instead of copying it in both places.

adapter_pipelines/Optimus/adapter.wdl
```diff
@@ -126,7 +126,7 @@ workflow AdapterOptimus {
   Int max_cromwell_retries = 0
   Boolean add_md5s = false
 
-  String pipeline_tools_version = "v0.46.0"
+  String pipeline_tools_version = "se-fix-timestamps"
```
pshapiro4broad: Is this a temporary change for testing that will be revered before merging up?

setup.py
samanehsan: This has definitely caused confusion in the past...but currently we use the `requirements.txt` to set-up our local development environment for pipeline-tools, and the `setup.py` requirements get used when the package is installed in the Dockerfile. They should be the same but it's harder to maintain when it's copied in two places.

adapter_pipelines/Optimus/adapter.wdl
samanehsan: Yes! When this code is run in the integration test, it will use the pipeline-tools docker image tagged with `se-fix-timestamps` so that the WDL tasks use the code from this PR.

samanehsan: Unfortunately there's room for error here because we have to remember to change it to the next version (`v0.47.0`) before merging. This is because the adapter WDLs are in the pipeline-tools repo, but also reference pipeline-tools docker images...so it is pretty complicated at the moment.

pshapiro4broad: Since they're workflow inputs it seems like there'd be a way to set these in an options file that's used for testing. But I guess you still need a way to communicate the docker tag required to the test runner.

samanehsan: That's true! We talked about doing that a while ago and haven't got around to it yet. But essentially Lira would need to include the version of `pipeline-tools` that it's using in the workflow inputs file that it generates dynamically before starting workflows.

samanehsan:
retest this please

samanehsan:
retest this please

PR #139, opened 2019-04-26
Refactor the whole file structure for future pipelines

rexwangcc:
### Purpose
_Please explain the purpose of this PR and include links to any GitHub issues that it fixes:_

The current structure is flat and encouraging developers mix everything together without thinking about the organization of functions, which has result in some ambiguity about what utils are responsible for which pipeline. Since there are more and more pipelines coming in the near future, we need a more reasonable way to organize the code so it's easier to find/change/extend.

#### Old structure:

```
pipelines_tools/
├── README.rst
├── __init__.py
├── auth_utils.py
├── confirm_submission.py
├── create_analysis_metadata.py
├── create_envelope.py
├── dcp_utils.py
├── gcs_utils.py
├── get_analysis_workflow_metadata.py
├── get_files_to_upload.py
├── get_upload_urn.py
├── http_requests.py
├── input_utils.py
├── optimus_utils.py
└── tests/
    └── data/
        └── metadata/
```

#### New structure:

```
pipelines_tools/
├── README.rst
├── __init__.py
├── pipelines/
│   ├── cellranger/
│   ├── optimus/
│   └── smartseq2/
├── shared/
│   └── submission/
└── tests/
    ├── data/
    │   └── metadata/
    ├── pipelines/
    │   ├── cellranger/
    │   ├── optimus/
    │   └── smartseq2/
    └── shared/
        └── submission/
```
---
### Changes
_Please list out what major changes were made in this PR to address the issue:_

**Commit 1:** Refactor the modules and files accoording to the discussion, fixed the imports, test would fail though.
**Commit 2:** Refactor the testing modules and files accoording to the discussion, fixed the imports, test should pass.

**Note: This PR doesn't change any code logically, it just moves things around and fix the imports and tests!!**

---
### Review Instructions
_Please provide instructions about how should a reviewer test/verify the changes in this PR:_

- No instructions.


ambrosejcarr:
It's often nice to have tests live in the packages they test (for example, a test folder in each of Optimus, ss2, cellranger), or: https://stackoverflow.com/questions/61151/where-do-the-python-unit-tests-go

Is there a particular reason you chose this structure?

rexwangcc:
@ambrosejcarr That's a good point. Organizing the tests folder in this way was intentional after a discussion with @samanehsan. There are 2 reasons:

1. Although it seems more convenient to have tests live in the packages they test, we found it more like a liability than a benefit to do so when the code base grows, especially if there are a lot of testing data that could be shared across tests. A good example is Angular Framework, which requires you to put tests next to the modules, you sometimes end up spending a lot of time just thinking about how to divide your test properly so that they can share the testing data while still living close to the src code. However, this is just a preference, not a strong argument.
2. More importantly, we are trying to avoid installing tests in site-packages (although quite a few distributions on PyPI are doing that) but keep them in the source distribution and could still be installed by something like `pip install blah[test]`. With this in mind, we chose the current hierarchy with a `MANIFEST.IN` file rather than something like:

```
pipelines_tools/
├── README.rst
├── __init__.py
├── pipelines/
│   ├── cellranger/
│                 └──tests/
│   ├── optimus/
│                 └──tests/
│   └── smartseq2/
│                 └──tests/
├── shared/
│   └──tests/
│   └── submission/
                     └──tests/
```

pipeline_tools/shared/metadata_utils.py
```diff
@@ -0,0 +1,141 @@
...
+        bundle (humancellatlas.data.metadata.Bundle): A Bundle object contains all of the necessary information.
+
+    Returns:
+        sample_id (str): String giving the sample id
+    """
+    sample_id = str(bundle.sequencing_input[0].document_id)
+    return sample_id
+
+
+def get_expected_cell_count(bundle):
```
samanehsan: This is only used by cellranger, so I'd say this can go in `cellranger.py`

rexwangcc: That's a good point, yeah we should move anything that's used by only one pipeline closer to that pipeline lib to prevent getting this file too large! Will do.

ambrosejcarr:
@rexwangcc Makes sense, particularly regarding (2) if you have substantial test data. 

> Although it seems more convenient to have tests live in the packages they test, we found it more like a liability than a benefit to do so when the code base grows, especially if there are a lot of testing data that could be shared across tests. A good example is Angular Framework, which requires you to put tests next to the modules, you sometimes end up spending a lot of time just thinking about how to divide your test properly so that they can share the testing data while still living close to the src code. However, this is just a preference, not a strong argument.

To mitigate this, we just have a `package/util/test` directory that contains our data and factories.

rexwangcc:
retest this please

rexwangcc:
retest this please

PR #147, opened 2019-05-22
Raise http error for 409 responses

samanehsan:
### Purpose
The input bundle information in the submission envelope is not getting saved because of a race condition in ingest that causes a 409 conflict error. On our side, we can retry the error but it requires a slight change to the retry logic so that an http error is raised for 409 status codes.

### Changes
Raise an error if a response status code is 409 so that tenacity retries the request

### Review Instructions

- No instructions.


samanehsan:
Not sure why the travis tests are not showing up now...they pass when I run them locally :octocat:

samanehsan:
retest this please

PR #149, opened 2019-06-05
Make it more tolerant on sequencing output formats

rexwangcc:
### Purpose

- Related discussions on Slack: https://humancellatlas.slack.com/archives/C9ENBPVNW/p1559590600019700
- Related issue in the metadata repos: https://github.com/HumanCellAtlas/metadata-schema/issues/971

### Changes

- Make it more tolerant on sequencing output formats

### Review Instructions

- No instructions.

### DIFF
--- a/pipeline_tools/pipelines/cellranger/cellranger.py
+++ b/pipeline_tools/pipelines/cellranger/cellranger.py
@@ -37,7 +37,9 @@ def create_cellranger_input_tsv(uuid, version, dss_url):
     # Parse inputs from metadata
     print('Gathering fastq inputs')
     fastq_files = [
-        f for f in primary_bundle.files.values() if f.file_format == 'fastq.gz'
+        f
+        for f in primary_bundle.files.values()
+        if f.file_format in ('fastq.gz', 'fastq')
     ]
     lane_to_fastqs = tenx_utils.create_fastq_dict(fastq_files)
 
--- a/pipeline_tools/pipelines/cellranger/cellranger.py
+++ b/pipeline_tools/pipelines/cellranger/cellranger.py
@@ -39,7 +39,7 @@ def create_cellranger_input_tsv(uuid, version, dss_url):
     fastq_files = [
         f
         for f in primary_bundle.files.values()
-        if f.file_format in ('fastq.gz', 'fastq')
+        if str(f.format).lower() in ('fastq.gz', 'fastq')
     ]
     lane_to_fastqs = tenx_utils.create_fastq_dict(fastq_files)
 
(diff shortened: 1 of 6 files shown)


pipeline_tools/pipelines/cellranger/cellranger.py
```diff
@@ -37,7 +37,9 @@ def create_cellranger_input_tsv(uuid, version, dss_url):
     # Parse inputs from metadata
     print('Gathering fastq inputs')
     fastq_files = [
-        f for f in primary_bundle.files.values() if f.file_format == 'fastq.gz'
+        f
+        for f in primary_bundle.files.values()
+        if f.file_format in ('fastq.gz', 'fastq')
```
malloryfreeberg: I don't know if this is necessary, but would doing something like 
```
if f.file_format.lower() in ('fastq.gz', 'fastq')
```
be useful to account for if a contributor puts "FASTQ" or "Fastq" for the format? I haven't specifically seen this before, but I'm trying to think ahead of the contributors :)

pipeline_tools/pipelines/optimus/optimus.py
```diff
@@ -27,7 +27,9 @@ def create_optimus_input_tsv(uuid, version, dss_url):
     # Parse inputs from metadata
     print('Gathering fastq inputs')
     fastq_files = [
-        f for f in primary_bundle.files.values() if f.file_format == 'fastq.gz'
+        f
+        for f in primary_bundle.files.values()
+        if f.file_format in ('fastq.gz', 'fastq')
```
malloryfreeberg: The file format field name is actually `format`.

https://github.com/HumanCellAtlas/metadata-schema/blob/e9acc562153d3e9941b91cea67f7828d1497ceef/json_schema/core/file/file_core.json#L31

Not sure if calling it `file_format` here is based on the schema, but if so, it should be `f.format` for re-ingested data :)

pipeline_tools/pipelines/cellranger/cellranger.py
rexwangcc: Sure, `.lower()` is defo safer. Will do!

pipeline_tools/pipelines/optimus/optimus.py
rexwangcc: I see, I will change it to `format`.

The reason why it's still working is because we are using @hannes-ucsc 's great lib here which shields us from the changes (https://github.com/HumanCellAtlas/metadata-api/blob/11480262abe484facebee016ce7e63a366c8743a/src/humancellatlas/data/metadata/api.py#L596)

I just checked our metadata version, it uses v6.0.0, which seems to be using the `format`: https://schema.dev.data.humancellatlas.org/core/file/6.0.0/file_core

malloryfreeberg: That's so great that the library worked!

hannes-ucsc: After property renames we introduce a wrapper for backwards compatibility. The wrapper is deprecated  so some IDEs show a warning when you use it. https://github.com/HumanCellAtlas/metadata-api/blob/develop/src/humancellatlas/data/metadata/api.py#L611

PR #150, opened 2019-06-10
Get 10x fastq file hashes

samanehsan:
### Purpose
In order for Lira to determine whether a bundle needs to be re-processed, it needs to have information about the fastq file hashes. 

### Changes
Since pipeline-tools already has functionality to parse fastq file metadata, update the `create_fastq_dict` function to return a mapping of read index files to their corresponding "manifest file entries", which contains properties such as the file url, sha256, sha1, etc: https://github.com/HumanCellAtlas/metadata-api/blob/develop/src/humancellatlas/data/metadata/api.py#L563


samanehsan:
Both test workflows passed the input-processing task:
- SS2: https://job-manager.caas-prod.broadinstitute.org/jobs/eb3d87d0-3055-4aee-89d6-efeb304f110a
- Optimus: https://job-manager.caas-prod.broadinstitute.org/jobs/bc280df8-533a-455c-801e-87a557f9e662

PR #151, opened 2019-06-27
GH-282: move max retries out of wdls and into config

ApChagi:
### Purpose

[JIRA](https://broadinstitute.atlassian.net/browse/GH-282)

### Changes

- max_retries runtime parameter moved out of wdl inputs and into cromwell options by way of Lira config
[parallel changes in lira](https://github.com/HumanCellAtlas/lira/pull/168)
[parallel changes in skylab](https://github.com/HumanCellAtlas/skylab/pull/226)

### Review Instructions

- No instructions.

### DIFF
--- a/adapter_pipelines/submit.wdl
+++ b/adapter_pipelines/submit.wdl
@@ -224,7 +224,7 @@ task confirm_submission {
   >>>
 
   runtime {
-    docker: "quay.io/humancellatlas/secondary-analysis-pipeline-tools:" + pipeline_tools_version,
+    docker: "quay.io/humancellatlas/secondary-analysis-pipeline-tools:" + pipeline_tools_version
   }
 
   output {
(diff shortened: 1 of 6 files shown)


samanehsan:
> Also, the specific case in GH-282 refers to maxRetries in Optimus, which currently doesn't have a max retries input, so -- as far as I can tell -- fixing that case doesn't require this change at all. It's unclear from the ticket if there's a requirement to update the cellranger and ss2 pipelines, which is what this change is doing.

@pshapiro4broad that is true, once we add maxRetries to the workflow options that is what will be used for Optimus. And it's less confusing to do the same thing for our other pipelines, hence removing `max_retries` as a workflow input to the adapter pipelines.

adapter_pipelines/submit.wdl
```diff
@@ -228,7 +224,7 @@ task confirm_submission {
   >>>
 
   runtime {
-    docker: "quay.io/humancellatlas/secondary-analysis-pipeline-tools:" + pipeline_tools_version
+    docker: "quay.io/humancellatlas/secondary-analysis-pipeline-tools:" + pipeline_tools_version,
```
rexwangcc: Is this `,` valid here?

samanehsan:
@ApChagi did you close this by mistake?

PR #155, opened 2019-07-03
Fix the failing tests.

rexwangcc:
### Purpose

- Fix the failing tests that were introduced in an earlier!

### Changes

- No changes.

### Review Instructions

- No instructions.

### DIFF
--- a/pipeline_tools/tests/shared/submission/test_confirm_submission.py
+++ b/pipeline_tools/tests/shared/submission/test_confirm_submission.py
@@ -2,9 +2,23 @@ import pytest
 import requests
 from tenacity import RetryError
 
+import unittest.mock as mock
 from pipeline_tools.shared.submission import confirm_submission
 from pipeline_tools.shared.http_requests import HttpRequests
 from pipeline_tools.tests.http_requests_manager import HttpRequestsManager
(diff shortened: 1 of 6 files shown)


pipeline_tools/tests/shared/submission/test_confirm_submission.py
```diff
@@ -68,50 +82,83 @@ def _request_callback(request, context):
...
 
-    def test_confirm_retries_on_error(self, requests_mock):
+    def test_confirm_retries_on_error(self, requests_mock, test_data):
         envelope_url = (
             'https://api.ingest.dev.data.humancellatlas.org/submissionEnvelopes/abcde'
         )
 
         def _request_callback(request, context):
             context.status_code = 500
+            assert request.headers.get('Authorization') == test_data.headers.get(
```
samanehsan: Since the test is mocking the return value of `DCPAuthClient.get_auth_header` I think this will always be true so I don't think you have to assert it here

samanehsan:
I have one comment about the tests, otherwise I think this is good to merge! :octocat:

samanehsan:
Going to merge this in now 🙂

PR #158, opened 2019-07-15
Make confirm submission step "retryable"

samanehsan:
### Purpose
The confirm submission task can fail in Cromwell after successfully making the request to confirm the submission. This leads to successful workflows having a failed status, which is difficult to debug. If this task used Cromwell's maxRetries feature, it could re-try the task in these situations. However, confirming an already confirmed submission throws an error, so this behavior needs to be handled first.

### Changes
1. When polling ingest for the submission envelope status, stop polling when the status goes "Valid", "Complete" or "Invalid". 
2. If the status is "Invalid", throw an error. This will allow us to "fail fast" and more easily differentiate timeout issues vs. actual errors with the submission envelope.
3. If the status is "Valid", confirm the submission.

### Review Instructions
~What if a submission is Invalid? If a task fails because the submission is invalid and the time-out limit was reached, the same thing will happen if the task is retried.~ Resolved with the above changes.

Note: After this is merged, the adapter workflow options need to specify a default `maxRetries` runtime parameter > 0 so that this task is retried by Cromwell if it fails

### DIFF
--- a/pipeline_tools/shared/submission/confirm_submission.py
+++ b/pipeline_tools/shared/submission/confirm_submission.py
@@ -4,7 +4,7 @@ import argparse
 from tenacity import retry_if_result, RetryError
 from datetime import datetime
 from pipeline_tools.shared.http_requests import HttpRequests
-from pipeline_tools.shared import auth_utils
+from pipeline_tools.shared import auth_utils, exceptions
 
 
 def wait_for_valid_status(envelope_url, http_requests):
@@ -102,7 +102,7 @@ def main():
         raise ValueError(message)
 
     if status == 'Invalid':
-        raise ValueError('Invalid submission envelope.')
+        raise exceptions.SubmissionError('Invalid submission envelope.')
     elif status == 'Valid':
         confirm(
             args.envelope_url,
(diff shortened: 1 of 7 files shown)


pipeline_tools/shared/submission/confirm_submission.py
```diff
@@ -94,17 +94,19 @@ def main():
     args = parser.parse_args()
     http_requests = HttpRequests()
     try:
-        wait_for_valid_status(args.envelope_url, http_requests)
+        response = wait_for_valid_status(args.envelope_url, http_requests)
```
aaclan-ebi: yeah, i believe this part will timeout if the submission is Invalid. Since the workflow is the owner of the submission, if it's not doing any correction to the metadata (I don't think there is). I think it's safe to fail early here if the submission goes invalid.

samanehsan:
retest this please

pipeline_tools/shared/submission/confirm_submission.py
```diff
@@ -94,17 +95,21 @@ def main():
...
         raise ValueError(message)
 
-    confirm(
-        args.envelope_url,
-        http_requests,
-        args.runtime_environment,
-        args.service_account_key_path,
-    )
+    if status == 'Invalid':
+        raise ValueError('Invalid submission envelope.')
```
rexwangcc: optional: raising a custom "SubmissionError" defined in `exceptions.py`

samanehsan:
FYI the integration test passed in Jenkins, but the status is not showing on the PR for some reason :octocat: https://gotc-jenkins.dsp-techops.broadinstitute.org/job/integration-tests/2425/console

PR #162, opened 2019-08-13
Remove data file checkout when only retrieving metadata

samanehsan:
### Purpose
When getting inputs to label workflows with a hash, pipeline-tools is also checking out the data files. This adds additional time to the request that is not necessary, since only the metadata is required at this stage.

### Changes
- Use the metadata-api library to download metadata
- Set `directurls` to `False` when retrieving metadata, but keep it set to `True` for the pipeline's input-processing step so that it can get gs urls to the data files.

### Review Instructions

- No instructions.


samanehsan:
Running this test workflow here: https://job-manager.caas-prod.broadinstitute.org/jobs/a6da7e8b-3026-413a-b87a-e9136df8c83d

samanehsan:
retest this please

samanehsan:
The integration test needs to use the "integration" environment instead of "test" now

rexwangcc:
@samanehsan That's a good point! I fixed it for the data-tracker but forgot the tests 🤦‍♂

PR #163, opened 2019-09-02
Fix getting bundle metadata in prod

samanehsan:
### Purpose
Fixes the dss deployment environment passed into the metadata-api `download_bundle_metadata` function in production, since it does not follow the same url pattern as the other environments.


samanehsan:
Going to merge in this bugfix since it is pretty urgent/fixes this funtion for prod

PR #164, opened 2019-09-05
Update file format mapping

samanehsan:
### Purpose
Update file format mapping used by pipeline submission step to remove the "unknown" file types from the metadata

### Changes
Added regex patterns to match file extensions `npy`, `npz` and `csv.gz` and updated the zarr file pattern to match all files in the zarr directory structure.

### DIFF
--- a/adapter_pipelines/format_map_example.json
+++ b/adapter_pipelines/format_map_example.json
@@ -13,6 +13,8 @@
   "[.]tsv$": "tsv",
   "[.]mtx$": "mtx",
   "[.]h5$": "h5",
-  "[.]csv$": "csv",
-  "[.]zattrs$": "matrix"
+  "[.]csv.gz$": "csv.gz",
+  "[.]zarr": "matrix",
+  "[.]npz$": "npz",
+  "[.]npy$": "npy"
 }


adapter_pipelines/format_map_example.json
```diff
@@ -13,6 +13,8 @@
   "[.]tsv$": "tsv",
   "[.]mtx$": "mtx",
```
rexwangcc: I'm not sure if it's ideal, but I remember someone mentioned `mtx` files should be considered as `matrix`. Maybe @HumanCellAtlas/pipelines-computational-biologists or @mckinsel could confirm?

samanehsan: Any thoughts on making both ".mtx" and "zarr" files display as "matrix" files in the data browser? @ambrosejcarr @barkasn @kishorikonwar @NoopDog

samanehsan: From @NoopDog: 
> My understanding is that currently the matrix label tags the existence of zarr files for the bundle. It means to us (rather loosely) that the matrix service “should” have this data.  Since matrix service is not always caught up with what has been ingested and any new pipelines (e.g. mouse data) then the matrix label is inaccurate. We are working with the matrix team to tighten up this interface.
Not sure if we should tag .mtx files the same way. We should discuss.

Given that this requires more discussion, I will merge this in as-is to fix the unknown file type issues

PR #165, opened 2019-09-06
deal with missing lane index

ApChagi:
### Purpose
Until the metadata team adds more information to allow grouping of files, the pipeline should error out when there are files that need to be grouped and insufficient information to do so. It should not error out when there is only one set of reads, and no need for grouping.

[JIRA](https://broadinstitute.atlassian.net/browse/GH-466)
[Related JIRA](https://broadinstitute.atlassian.net/browse/GH-173)

### Changes

- No changes.

### Review Instructions

- No instructions.


samanehsan:
Looks good!

PR #170, opened 2019-10-07
Update reference paths to use prod bucket

samanehsan:
### Purpose
The reference files are split between different buckets in our dev google project. This also has the risk that someone may think the files are only used for testing and move/delete them 🙀 

### Changes
Copy the reference files into the `hca-dcp-analysis-pipelines-reference` bucket for production and update all of the paths to use those files

### Review Instructions
Any suggestions on file structure? The only thing I changed was the bucket, not the actual reference folder name/structure


samanehsan:
retest this please

PR #171, opened 2019-10-08
Add format_map to pipeline_tools code

samanehsan:
### Purpose
Use the file_format map defined in the code, instead of using a copy stored in gcloud. 

### Changes

- No changes.

### Review Instructions

- No instructions.


pipeline_tools/shared/submission/format_map.json
```diff
@@ -0,0 +1,21 @@
...
+  "[.]bai$": "bai",
+  "[.]tsv$": "tsv",
+  "[.]mtx$": "mtx",
+  "[.]h5$": "h5",
+  "[.]csv$": "csv",
+  "[.]csv.gz$": "csv.gz",
+  "[.]zarr": "matrix",
+  "[.]npz$": "npz",
+  "[.]npy$": "npy"
+}
```
tbl3rd: Do I sense a pattern?

samanehsan: Yes 😆

samanehsan:
retest this please

pipeline_tools/shared/submission/create_analysis_metadata.py
```diff
@@ -511,18 +511,13 @@ def main():
...
-        help='JSON file providing map of file extensions to formats.',
-    )
     parser.add_argument(
         '--add_md5s', help='Set to "true" to add md5 checksums to file metadata'
     )
     args = parser.parse_args()
 
     # Get the extension_to_format mapping
-    with open(args.format_map) as f:
+    with open('format_map.json') as f:
```
rexwangcc:
```suggestion
    with open('format_map.json', 'r') as f:
```

rexwangcc:
Based on the offline discussion, I'm a bit concerned that the JSON might not always be accessible to the python package. One alternative could be having a separate "format_map.py" module in the shared folder and define the loaded dict inside so we don't have to load the JSON. But if both work, it doesn't matter which one you'd like to use!

PR #172, opened 2019-11-18
Get Optimus chemistry from bundle metadata

samanehsan:
### Purpose
The newest version of Optimus requires a "chemistry" parameter to determine whether the data is for a 10x v2 or v3 assay, so the adapter pipeline must provide this info to the pipeline. 

### Changes
Get the library construction method ontology from the metadata and use that to determine whether the Optimus chemistry. Then, write the value to a file called "chemistry.txt" so that it can be included with the rest of the outputs of the adapter WDL's prep task.

Note: Right now, this only supports Optimus v2 chemistry to simplify the PR. I will open up a follow-up PR that handles Optimus v3 and updates the input hashing function.

### Review Instructions
- No instructions.

### DIFF
--- a/pipeline_tools/pipelines/optimus/chemistry.py
+++ b/pipeline_tools/pipelines/optimus/chemistry.py
@@ -14,7 +14,7 @@ class Chemistry(Enum):
 class LibraryConstructionMethod(Enum):
 
     """
-    This class maps to the 10x library construction ontology ids:
+    This class maps to library construction ontology ids, e.g:
     https://www.ebi.ac.uk/ols/ontologies/efo/terms?iri=http%3A%2F%2Fwww.ebi.ac.uk%2Fefo%2FEFO_0008995&viewMode=All&siblings=false
     """
 
--- a/pipeline_tools/pipelines/optimus/chemistry.py
+++ b/pipeline_tools/pipelines/optimus/chemistry.py
@@ -20,7 +20,6 @@ class LibraryConstructionMethod(Enum):
 
     tenX_v2 = "EFO:0009310"
     tenX_3_prime_v2 = "EFO:0009899"
-    tenX_5_prime_v2 = "EFO:0009900"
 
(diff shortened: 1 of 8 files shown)


pipeline_tools/pipelines/optimus/chemistry.py
```diff
@@ -0,0 +1,27 @@
...
+class LibraryConstructionMethod(Enum):
+
+    """
+    This class maps to library construction ontology ids, e.g:
+    https://www.ebi.ac.uk/ols/ontologies/efo/terms?iri=http%3A%2F%2Fwww.ebi.ac.uk%2Fefo%2FEFO_0008995&viewMode=All&siblings=false
+    """
+
+    tenX_v2 = "EFO:0009310"
+    tenX_3_prime_v2 = "EFO:0009899"
+    tenX_5_prime_v2 = "EFO:0009900"
```
rexwangcc: We might want to remove this to avoid confusion in the future since it's not on the roadmap

pipeline_tools/pipelines/optimus/optimus.py
```diff
@@ -16,6 +20,36 @@
     },
 }
 
+LIBRARY_CONSTRUCTION_METHODS = {
+    Chemistry.tenX_v2.value: [
+        LibraryConstructionMethod.tenX_v2.value,
+        LibraryConstructionMethod.tenX_3_prime_v2.value,
+        LibraryConstructionMethod.tenX_5_prime_v2.value,
```
rexwangcc: And here as well

samanehsan:
retest this please

samanehsan:
Here is the mintegration test workflow that I ran for Optimus using v1.4.0, this branch of pipeline-tools and the corresponding branch for the adapter-pipeline: https://job-manager.caas-prod.broadinstitute.org/jobs/0555fc3d-23b0-4269-9a52-60a9fa457e36

PR #200, opened 2021-08-20
add reference file to copy script, also make it so we can take in mul…

nikellepetrillo:
add metadata and descriptor reference files to copy script.
make it so we can take in multiple output_file_paths in create_links.py

### Purpose

- No issue is linked to this PR.

### Changes

- No changes.

### Review Instructions

- No instructions.


pipeline_tools/shared/submission/create_links.py
```diff
@@ -191,7 +191,7 @@ def main():
     parser.add_argument('--analysis_protocol_path', required=True, help='Path to the /metadata/analysis_protocol.json file.')
     parser.add_argument('--project_level', type=bool, default=False, required=False, help='Boolean representing project level vs intermediate level.')
     parser.add_argument('--workspace_version', required=True, help='A version (or timestamp) attribute shared across all workflows''within an individual workspace.')
-    parser.add_argument('--output_file_path', required=False, help='Path to the outputs.json file (This is just a json list of the /metadata/analysis_file/*.json files).')
+    parser.add_argument('--output_file_path', required=False, nargs='+', help='Path to the outputs.json file (This is just a json list of the /metadata/analysis_file/*.json files).')
```
wjdingman1, line 194: is there a case where there will be multiple outputs.json files? Don't think we need nargs here

PR #225, opened 2021-09-07
Removing unnecessary capitalization for pipeline_version

cheyenne-gold:
also fixing error for optimus. seeing `UnboundLocalError: local variable 'single_sample_pipeline_version' referenced before assignment` because that variable is never defined for optimus. updated check for the string version of the variable name in the script locals.


jessicaway (approved):
looks good to me

